PR #6, opened 2025-12-17
feat: add on-demand model loading option for transcription

materemias:
Model loading is swift, taking under a second on modern systems. This new approach optimizes VRAM usage by loading the model only at recording start and unloading it upon transcription completion.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -42,6 +42,11 @@ translate = false
 # Number of CPU threads for inference (omit for auto-detection)
 # threads = 4
 
+# Load model on-demand when recording starts (true) or keep loaded (false)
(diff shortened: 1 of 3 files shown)


peteonrails:
This is a really good idea. I do not have a VRAM constrained system, so I did not consider this optimization. I like it. I would like to change the default value to "false" and make the on-demand loading be opt-in for now. 

I am concerned about one possible race condition when a user keys the hotkey and unkeys quickly before the model loads - this might cause a resource leak. 

I'm also thinking about how we might better handle the panic call - I've pulled your forked branch and am adding a few commits on top to this effect. 

THANK YOU so much for the contribution. I've added you to the contributor list in the README and on the website.

PR #12, opened 2025-12-22
Set language on mod key

materemias:
This feature allows you to assign language profiles to modifier key combinations (for example, RCTRL+SCR_LCK to "en", RSHIFT+SCR_LCK to "it"), while preserving the original modifier key behavior. The goal is to enhance transcription accuracy by providing a cue for the spoken language.


peteonrails:
Closed as part of the git history rewrite, but this idea will be reconsidered soon

PR #43, opened 2026-01-08
Feat/nixos packaging

peteonrails:
## Description

Experimental support for NixOS packaging. Initial spike - builds in a container both GPU and CPU. Needs more testing from folks with NixOS as their daily driver.


nix/nixos-module.nix
```diff
@@ -0,0 +1,127 @@
...
+        users could potentially act as a keylogger. Only use this if:
+
+        - You need the built-in evdev hotkey (most users should use compositor
+          keybindings instead, which don't require special permissions)
+        - You understand and accept the security implications
+        - You're on a single-user system with trusted software
+
+        Leave empty (default) if using compositor keybindings for push-to-talk.
+      '';
+    };
```
danheuck: I'm not sure whether this is a problem, but I don't think I've seen any other packages follow this pattern. Normally they just tell you what group you need to be part of and expect you to add it in your `users.users.<user>.extraGroups` yourself.

A related thought, given the security warnings throughout this flake about adding your user to the input group, is it possible for voxtype to run as it's own user and to give that user the input group? Not sure if that would break typing or something

nix/nixos-module.nix
```diff
@@ -0,0 +1,127 @@
...
+      partOf = [ "graphical-session.target" ];
+      after = [ "graphical-session.target" ];
+
+      serviceConfig = {
+        ExecStart = "${pkgs.ydotool}/bin/ydotoold";
+        Restart = "on-failure";
+        RestartSec = 5;
+      };
+    };
+  };
```
danheuck: nixos has a ydotool module `programs.ydotool`, seems like we should just use that instead of making our service for it. We do need to make sure the user is in the `programs.ydotool.group` group for that one though.

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/ydotool.nix

danheuck (commented):
Left some comments. I'm not experienced enough with nix packaging to be authoritative here, so take them or leave them. Overall it looks pretty good, I just think it can be simplified a bit. I haven't tested the actual nixos/home-manager modules yet since my own minimal hm module is working fine at the moment; I'll update if/when I do.

I had a thought while looking at all of those security warnings for adding the user to the `input` group: is it possible for voxtype to run as it's own separate user and to give that user the input group? Not sure if that would work since voxtype needs to emulate input, but it would minimize the security risk.

The first time I ran the build it failed a test. Looking at the test, I guess it timed out? Not sure why as it sure doesn't seem like it should've taken 5 seconds. 
Subsequent builds of the default and vulkan packages built fine, but I thought I'd flag it.

```shell
❯ nix build github:peteonrails/voxtype/feat/nixos-packaging#default
error: Cannot build '/nix/store/l5n15684hyg0baas1i5jwghyn61glsj7-voxtype-0.4.9.drv'.
       Reason: builder failed with exit code 101.
       Output paths:
         /nix/store/1njir7xhhgankczqvwj6wvahh7yaygmk-voxtype-0.4.9
       Last 25 log lines:
       > test text::tests::test_newline_and_tab ... ok
       > test text::tests::test_spoken_punctuation_case_insensitive ... ok
       > test output::post_process::tests::test_sed_transformation ... ok
       > test text::tests::test_spoken_punctuation_multi_word ... ok
       > test text::tests::test_spoken_punctuation_basic ... ok
       > test text::tests::test_developer_punctuation ... ok
       > test output::post_process::tests::test_timeout_fallback ... ok
       >
       > failures:
       >
       > ---- output::post_process::tests::test_whitespace_trimming stdout ----
       >
       > thread 'output::post_process::tests::test_whitespace_trimming' (2581) panicked at src/output/post_process.rs:238:9:
       > assertion `left == right` failed
       >   left: "ignored"
       >  right: "hello"
       > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
       >
       >
       > failures:
       >     output::post_process::tests::test_whitespace_trimming
       >
       > test result: FAILED. 70 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
       >
       > error: test failed, to rerun pass `--lib`
       For full logs, run:
         nix log /nix/store/l5n15684hyg0baas1i5jwghyn61glsj7-voxtype-0.4.9.drv
error: Cannot build '/nix/store/9wpdnj0djw6dxf667wlddiakbpxsxgnf-voxtype-wrapped-0.4.9.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/pk1krqrv9vjza66dbnkgkigr5mmra4f8-voxtype-wrapped-0.4.9
```

PR #63, opened 2026-01-10
fix: add wtype init delay to prevent first character drop

### LINKED ISSUE #61: [Bug] First character of transcription dropped when using compositor submap integration
## Description

After setting up the Hyprland compositor integration via `voxtype setup compositor hyprland`, the first character of transcriptions is intermittently dropped/not typed.

For example, "Okay" becomes "kay", "Hello" becomes "ello".

This does not happen 100% of the time, but frequently enough to be noticeable.

## Environment

- **VoxType version:** 0.4.11
- **OS:** Arch Linux (kernel 6.18.3-arch1-1)
- **Hyprland version:** 0.53.1
- **wtype version:** 0.4-2
- **Display server:** Wayland

## Configuration

Using remote whisper backend pointing to a local OpenAI-compatible server.

`~/.config/voxtype/config.toml` (relevant sections):
```toml
[whisper]
backend = "remote"
remote_endpoint = "http://localhost:8777"
remote_timeout_secs = 180
language = "en"

[output]
mode = "type"
pre_output_command = "hyprctl dispatch submap voxtype_suppress"
post_output_command = "hyprctl dispatch submap reset"
fallback_to_clipboard = true
type_delay_ms = 10
```

Hyprland submap config (`~/.config/hypr/conf.d/voxtype-submap.conf`) was generated by `voxtype setup compositor hyprland`.

## Steps to Reproduce

1. Install voxtype 0.4.11
2. Run `voxtype setup compositor hyprland`
3. Add `source = ~/.config/hypr/conf.d/*.conf` to `~/.config/hypr/hyprland.conf`
4. Reload Hyprland: `hyprctl reload`
5. Start voxtype daemon
6. Use push-to-talk hotkey (e.g., SUPER+SHIFT+D) to record and transcribe
7. Observe that the first character is sometimes missing

## Attempted Workarounds

- Increased `type_delay_ms` from 1 to 10 - no effect
- Added sleep to pre_output_command: `hyprctl dispatch submap voxtype_suppress && sleep 0.3` - no effect

## Suspected Cause

Possible race condition between the submap activation and wtype starting to send keystrokes. The first keystroke may be sent before Hyprland fully switches to the `voxtype_suppress` submap.

## Notes

- The compositor integration itself works correctly - Walker (app launcher bound to SUPER+SPACE) no longer opens during transcription
- Issue appears specific to the submap integration; may not occur without `pre_output_command`

Julian194:
## Summary

Adds a 200ms delay (`-s 200`) after wtype initializes its virtual keyboard before sending the first keystroke. This fixes an issue where the first character of transcriptions was intermittently dropped.

## Root Cause

When wtype starts, it connects to Wayland and creates a virtual keyboard. The first keystroke was being sent before the virtual keyboard was fully ready, causing the first character to be dropped.

## Testing

Tested on Arch Linux with Hyprland. Before this fix, transcriptions like "Hello" would appear as "ello". After the fix, all characters appear correctly.

## Note

This fix works best when combined with disabling the submap integration (commenting out `pre_output_command` and `post_output_command` in config). The submap was also contributing to the issue by consuming virtual keyboard events.

Fixes #61

### DIFF
--- a/src/output/wtype.rs
+++ b/src/output/wtype.rs
@@ -63,6 +63,7 @@ impl TextOutput for WtypeOutput {
         }
 
         let output = Command::new("wtype")
+            .args(["-s", "200"]) // delay after init to prevent first char drop
             .arg("--")
             .arg(text)
             .stdout(Stdio::null())


peteonrails:
I am going to make this a configurable value instead of the hardcoded 200ms and credit you as coauthor on the commit. Thanks so much for your help  troubleshooting this!

PR #65, opened 2026-01-10
Bunch of optimizations and random semi baked features to let me use this

lypanov:
I have no expectation whatsoever that you will accept this as is, feel free to ignore, however I can heavily recommend at least pulling in the first optimization. Unfortunately right now I don't have time to continue with it, given that it's working well enough for my use case. As soon as I do have time, I'll try to polish the patches up. The only really critical one anyway is Optimization 1, which is a trivial patch which adds the usage of audio context. This massively improves the time taken to transcribe short sentences, especially for CPU-only users. The cut off point at 75% of the batch window (30s) is entirely arbitrary, I just feel like it's a good amount given that there's a feasibility that audio context can actually reduce quality (I haven't seen this in my own personal testing up to now, but you never know).

Also, it's entirely vibe coded as I only have basic knowledge of Rust.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -102,13 +102,16 @@ pub enum OutputError {
     #[error("wl-copy not found in PATH. Install wl-clipboard via your package manager.")]
     WlCopyNotFound,
 
+    #[error("xclip not found in PATH. Install via your package manager.")]
+    XclipNotFound,
+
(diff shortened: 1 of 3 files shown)


peteonrails:
No worries on timing - I'm here to help when you're ready

PR #86, opened 2026-01-15
Fix paste mode keystroke timing with ydotool

materemias:
## Summary
- Add configurable delay for ydotool paste keystroke using `-d` parameter
- Use `type_delay_ms` config value to control timing between key events
- Only pass delay parameter when explicitly configured (non-zero)

## Problem
The paste mode was experiencing intermittent failures with ydotool because key events were being processed too fast without proper timing. This caused:
- Intermittent paste failures (paste keystroke not being recognized)
- Capital "V" appearing instead of the paste command executing

The issue was that Ctrl was being released before the V keystroke was fully registered, breaking the Ctrl+V combination.

## Solution
Added support for ydotool's `-d` parameter to introduce configurable delays between key events. Users can now adjust timing via the `type_delay_ms` config value to match their system's keyboard event processing speed.

When `type_delay_ms = 0` (default), no delay parameter is passed and ydotool uses its defaults.

## Test Plan
- [x] Verify paste mode works with default `type_delay_ms = 0`
- [x] Test with increased `type_delay_ms` values (e.g., 50-100ms)
- [x] Confirm keystroke combinations work (ctrl+v, shift+insert, ctrl+shift+v)
- [x] Verify no capital "V" appears when using ydotool fallback

### DIFF
--- a/src/output/paste.rs
+++ b/src/output/paste.rs
@@ -456,8 +456,7 @@ impl TextOutput for PasteOutput {
         self.copy_to_clipboard(text).await?;
 
         // Small delay to ensure clipboard is set before pasting
-        // Increased from 100ms to 200ms to improve reliability with ydotool
-        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
 
         // Step 2: Simulate paste keystroke
         self.simulate_paste_keystroke().await?;


src/output/paste.rs
```diff
@@ -446,7 +456,8 @@ impl TextOutput for PasteOutput {
         self.copy_to_clipboard(text).await?;
 
         // Small delay to ensure clipboard is set before pasting
-        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        // Increased from 100ms to 200ms to improve reliability with ydotool
+        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
```
peteonrails: Might be time to make this a configurable parameter too. What do you think?

materemias: I think this leads to config fatigue, as long as it fits most users (no related issue report) I would leave it this way

PR #90, opened 2026-01-15
Add single-instance safeguard using pidlock

materemias:
## Summary
Prevents multiple voxtype daemon instances from running simultaneously by acquiring a PID lock at startup. If another instance is already running, the daemon exits gracefully with a clear error message.

The lock is maintained for the entire daemon lifetime and automatically released on shutdown, preventing conflicts with audio capture, hotkey binding, and output handling.

## Changes
- Added `pidlock` dependency to Cargo.toml
- Imported `Pidlock` in daemon.rs
- Added single-instance check in `Daemon::run()` method that acquires a lock at `$XDG_RUNTIME_DIR/voxtype/voxtype.lock`

## Test plan
- [x] Code compiles without errors
- [x] All existing tests pass (124 tests)
- [x] Verify lock file is created at startup and released on shutdown

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1161,6 +1161,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
+name = "pidlock"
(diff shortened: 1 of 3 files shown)


peteonrails:
@materemias This is an excellent addition. THANK YOU

PR #140, opened 2026-01-28
flake: configure onnxruntime package

andresilva:
Both the build inputs and the runtime wrapper were using the default CPU-only `onnxruntime` for all parakeet variants. Now the CUDA and ROCm builds use `onnxruntime.override { cudaSupport = true; }` and `onnxruntime.override { rocmSupport = true; }` respectively.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -60,17 +60,21 @@
         # Wrap a parakeet package with runtime dependencies and ORT_DYLIB_PATH
         # Parakeet uses ONNX Runtime, which needs to know where to find the library
         libExt = if pkgs.stdenv.isDarwin then "dylib" else "so";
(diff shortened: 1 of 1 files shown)


peteonrails (approved):
LGTM. This correctly configures GPU-enabled ONNX Runtime for the CUDA and ROCm Parakeet variants.

Key points verified:
- Uses `pkgsUnfree` for CUDA variant (CUDA has unfree license)
- Uses `pkgs` for ROCm variant (ROCm is open source)
- Both build inputs and runtime wrapper use the matching ONNX Runtime variant
- CPU-only `parakeet` package still uses default ONNX Runtime

Thanks for the fix!

PR #147, opened 2026-01-29
WIP: VAD with Whisper and Silero backends

peteonrails:
## Summary

Implements Voice Activity Detection with engine-specific backends:

- **WhisperVad**: Uses whisper-rs built-in VAD (GGML Silero model) - works now
- **SileroVad**: Uses `voice_activity_detector` crate for Parakeet builds - blocked on ort compatibility

## Status: Blocked

The `voice_activity_detector` 0.2.0 crate uses an older `ort` API that conflicts with `parakeet-rs`'s ort version. Build fails with API mismatch errors when the `parakeet` feature is enabled.

## Unblocking options

1. Wait for `voice_activity_detector` to update for newer ort versions
2. Find an alternative Silero ONNX implementation compatible with our ort version
3. Contribute ort compatibility fix upstream to `voice_activity_detector`

## Related

- See `feature/vad-silence-filter` for the working energy-based VAD (no external dependencies)
- WhisperVad from this branch can be cherry-picked independently since it has no ort dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -741,10 +741,93 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
+name = "futures"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
+dependencies = [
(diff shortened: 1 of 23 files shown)


peteonrails:
## Update: ort compatibility may be unblocked

The `silero-vad-rust` crate (v6.2.1) uses `ort 2.0.0-rc.10`, which is compatible with our `ort 2.0.0-rc.11`. This is a different crate from `voice_activity_detector` and doesn't have the API conflict.

This could replace the blocked `voice_activity_detector` dependency and give us Silero VAD in ONNX binaries. Worth spiking as a replacement.

Issue #47, opened 2026-01-08
[Feature] Nemotron Speech backend


peteonrails:
Explore using NVIDIA Nemotron ASR model as a backend.

lokkju:
see PR #155  - working well, but CPU performance is a bit slow; quants may help.

PR #155, opened 2026-01-30
Nemotron streaming transcription (real-time text output)

lokkju:
## Summary

Adds Nemotron Speech Streaming EN 0.6B support (issue #47) to the Parakeet engine, enabling real-time text output during recording. Text is typed incrementally as you speak rather than waiting until recording stops.

- **New model type**: `ParakeetModelType::Nemotron`, auto-detected from model files
- **New trait**: `StreamingTranscriber` for incremental chunk-based transcription
- **New state**: `State::StreamingRecording` with live text output during recording
- **Persistent model**: loaded once at daemon startup, reused across recordings
- **Model download**: available via `voxtype setup model` (option 13)
- **parakeet-rs bump**: 0.2.9 → 0.3.1 (adds Nemotron API, renames `rocm` → `migraphx`)

### How it works

Audio chunks (560ms / 8960 samples at 16kHz) are fed to the Nemotron model via a persistent blocking task. Text deltas are sent back through a channel and typed immediately via the output chain. On recording stop, remaining audio is flushed through silence padding.

### Config

```toml
engine = "parakeet"

[parakeet]
model = "nemotron-speech-streaming-en-0.6b"
# streaming is auto-enabled for Nemotron models
# streaming = false  # to force batch mode instead
```

## Status: Not Ready

This branch is functional but has a significant limitation:

**Inference speed on CPU is too slow for real-time use.** The 0.6B fp32 model takes longer than 560ms to process each chunk on CPU, so text output lags behind speech. It works, but with noticeable delay.

### Path forward

- **int8/int4 quantized models** are being worked on. Dynamic int8 quantization should bring the model from ~2.6 GB to ~670 MB with proportionally faster inference. int4 would be even smaller/faster. This is the most likely fix for CPU users.
- **CUDA acceleration** (`--features parakeet,parakeet-cuda`) would make this real-time on NVIDIA GPUs but hasn't been tested yet.
- **MIGraphX** (`--features parakeet,parakeet-rocm`) for AMD discrete GPUs — compiles but untested.

## Test plan

- [x] `cargo build --features parakeet` compiles
- [x] `cargo build` without parakeet compiles (no regressions)
- [x] `cargo test` — 282 tests pass
- [x] Manual: Nemotron model appears in `voxtype setup model`
- [x] Manual: text appears live during recording
- [x] Manual: second recording works without errors
- [x] Manual: batch transcription still works for CTC/TDT models
- [ ] Performance: real-time streaming on CPU with quantized model
- [ ] Test CUDA acceleration

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,9 +1594,9 @@ dependencies = [
 
 [[package]]
 name = "parakeet-rs"
-version = "0.2.9"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a7667842fd2f3b97b029a30fb9a00138867c6915229f5acd6bd809d08250d2ee"
+checksum = "447e20c2f0ecc35e581aa30da451e4238221a0dfd020c4ee6bdba7cb8c7b52dd"
 dependencies = [
  "eyre",
  "hound",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ rodio = { version = "0.19", default-features = false, features = ["wav"] }
 whisper-rs = "0.15.1"
 
 # Parakeet speech-to-text (optional, ONNX-based)
-parakeet-rs = { version = "0.2.9", optional = true }
+parakeet-rs = { version = "0.3.1", optional = true }
 
 
 # CPU count for thread detection
@@ -89,7 +89,7 @@ gpu-hipblas = ["whisper-rs/hipblas"]
 parakeet = ["dep:parakeet-rs"]
 parakeet-cuda = ["parakeet", "parakeet-rs/cuda"]
 parakeet-tensorrt = ["parakeet", "parakeet-rs/tensorrt"]
-parakeet-rocm = ["parakeet", "parakeet-rs/rocm"]
+parakeet-rocm = ["parakeet", "parakeet-rs/migraphx"]
 # Dynamic loading for system ONNX Runtime (used by Nix builds)
 parakeet-load-dynamic = ["parakeet", "parakeet-rs/load-dynamic"]
 
--- a/contrib/nemotron-streaming-test-config.toml
+++ b/contrib/nemotron-streaming-test-config.toml
@@ -1,0 +1,46 @@
+# Voxtype test config: Nemotron streaming transcription
+#
+# Setup:
+#   1. Build with parakeet support:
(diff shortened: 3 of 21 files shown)


lokkju:
I'm not sure how the modifier key sequences would affect output? at least when using wayland's virtual keyboard protocol or libei compat such as in Mutter, it should be treated as a seperate input device, so the modifier keys on your actual keyboard shouldn't affect the input coming from wtype/eitype; is that what you were potentially seeing as a problem?

My main daily driver is a Framework 13, and it's F12 function key is a gear icon that maps to KEY_MEDIA; absolutely perfect for the hold-to-talk key.

Right now, with the int4 quant, it's still just slightly too slow for my preferences; but you do get the typing as you speak, which is quite nice. I'm exploring the webgpu backend, but so far it's limited featureset seems to actually cause slower processing than avx512. Have you tried with ncnn at all? supposedly ONNX can be converted, and it has great vulkan backend support.

peteonrails:
@lokkju 

> so the modifier keys on your actual keyboard shouldn't affect the input coming from wtype/eitype; is that what you were potentially seeing as a problem?

Yes, when Omarchy incpororated Voxtype, users would report seeing wild results when unkeying SUPER-SHIFT-X because the keystrokes started before the user finished unkeying - I want to make sure I don't set folks up for that. 

But if you are right about it being a separate input device, then it sounds like the diagnosis may have been off - I'll have to look in to that some more. 

Have not tried ncnn but will allocate a little time to look in to it. Looking forward to getting nemotron merged soon - even if experimental!

peteonrails:
@lokkju 

> It's looking like the webgpu experiments won't help much here, nor vulkan backends;

That is unfortunate - I  hope to get some numbers for you on my testing workstation this weekend: here's hoping they show some promise

lokkju:
So at this point, it's not the models; it's the rest of the system around it. I've optimized the typing tool check to cache on launch, and that saves 5ms twice per second, which ads up. similar small gains elsewhere as well. I think the next step is to use eitype/wrtype as a library rather than launch as a tool; I'll be trying that tonight or tomorrow. Regardless, it's usable right now, and a better experience than typing for a while, then waiting for the whole transcription, for sure.

peteonrails:
@lokkju Thanks for all of your work on this - going to close this approach for now in favor of picking back up later under the newer streaming architecture work we've done

PR #164, opened 2026-02-01
Fix CLI Description of recording commands 

barrettruth:
The help text showed hyphenated commands (record-toggle) but the CLI uses subcommands with spaces (record toggle) - simple update the cli to reflect this..


peteonrails:
Thanks for this documentation fix, @barrettruth , much appreciated! Any chance you could sign your commit?

barrettruth:
oh yeah, no idea how that happened.

PR #265, opened 2026-03-16
Add post-processing to meeting mode and dictation context

materemias:
## Summary

This PR adds post-processing support to two areas that previously lacked it:

### 1. Meeting mode post-processing (new)

Meeting mode previously skipped post-processing entirely. Now, if `[output.post_process]` is configured, each transcribed chunk is piped through the post-processing command before being stored in the transcript. Previous chunk text is passed as context so the LLM can maintain continuity across 30-second chunk boundaries.

Context is tracked per audio source (mic vs loopback) to prevent speaker bleed between channels. Context is cleared when a meeting stops so it doesn't leak into the next meeting.

### 2. Dictation mode context (new)

Regular push-to-talk dictation now passes the previous dictation's text as context to the post-processing command, if the previous dictation was within 60 seconds. This helps the LLM produce more coherent cleanup when dictating in quick succession.

The 60-second retention window is intentionally not configurable to avoid config fatigue. It covers rapid-fire dictation without leaking stale context.

### Context delivery mechanism

Context is passed via the `VOXTYPE_CONTEXT` environment variable, keeping stdin as plain current text. This means existing post-processing scripts work unchanged. Scripts that want context can optionally read `$VOXTYPE_CONTEXT`.

The env var is explicitly cleared before each command invocation to prevent inheriting stale context from the parent environment.

### Other changes

- `process()` now delegates to `process_with_context(text, None)`, removing ~20 lines of duplicated error-handling logic
- Post-processing tracing logs now include `changed: true/false` to indicate whether the command modified the text
- Example scripts (gemini, ollama, openai) updated to optionally use `$VOXTYPE_CONTEXT`
- Docs updated (CONFIGURATION.md, USER_MANUAL.md)

## Test plan

- [ ] Configure `[output.post_process]` and run a meeting, verify exported transcript shows post-processed text
- [ ] Verify post-processing failure falls back to original text
- [ ] Verify mic and loopback chunks get independent context (no speaker bleed)
- [ ] Verify context clears on meeting stop and doesn't leak to next meeting
- [ ] Verify regular dictation gets context from previous dictation within 60s
- [ ] Verify dictation context expires after 60s
- [ ] Verify existing scripts work without modification (stdin is plain text)
- [ ] Verify `VOXTYPE_CONTEXT` is not inherited from parent environment

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,26 +155,6 @@ checksum = "7d809780667f4410e7c41b07f52439b94d2bdf8528eeedc287fa38d3b7f95d82"
 
 [[package]]
 name = "bindgen"
-version = "0.71.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f58bf3d7db68cfbac37cfc485a8d711e87e064c3d0fe0435b92f7a407f9d6b3"
-dependencies = [
- "bitflags 2.10.0",
- "cexpr",
- "clang-sys",
- "itertools 0.13.0",
- "log",
- "prettyplease",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash",
- "shlex",
- "syn",
-]
-
-[[package]]
-name = "bindgen"
 version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


peteonrails:
Thanks @materemias - I'll spend some time testing these changes. They look good so far. I appreciate your contribution!

PR #329, opened 2026-04-23
fix: preserve remote STT model selection

krystophny:
## Summary

This PR is intentionally minimal. Voxtype already supports remote/OpenAI-compatible Whisper endpoints via `[whisper] backend = "remote"` or `mode = "remote"`.

The only change here fixes remote mode inside `ModelManager`: remote API model names are endpoint-specific and must not be validated against or replaced by local whisper.cpp model names.

## Fix

- Preserve configured `remote_model` when remote mode is used without a per-request model override.
- Allow explicit remote model overrides to pass through without local model availability checks.
- Skip local model preparation for remote mode.
- Add a focused test proving the remote request sends `whisper-1` rather than the local default model name such as `base.en`.

## Verification

Executed on `mailuefterl` in `/tmp/voxtype-client-min.OO0FyN/repo`:

```text
$ cargo test test_remote_backend_uses_remote_model_not_local_model -- --nocapture
1 passed; 0 failed
```

```text
$ cargo check
Finished `dev` profile
```

```text
$ cargo build
Finished `dev` profile
```

```text
$ cargo test
540 unit tests passed; 25 integration tests passed; 0 failed
```

### DIFF
--- a/src/model_manager.rs
+++ b/src/model_manager.rs
@@ -68,6 +68,12 @@ impl ModelManager {
         &mut self,
         model: Option<&str>,
     ) -> Result<Arc<dyn Transcriber>, TranscribeError> {
+        // Remote API models are endpoint-specific and should not be validated
+        // against local whisper.cpp model names.
+        if self.config.effective_mode() == WhisperMode::Remote {
+            return self.create_remote_transcriber(model);
+        }
+
         // Clone the model name to avoid borrow issues
         let model_name = model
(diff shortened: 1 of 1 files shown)


krystophny:
Closing this draft as redundant: Voxtype already supports OpenAI-compatible remote Whisper endpoints via explicit remote backend config. For local testing, we can configure the existing remote backend instead of carrying a code change.

PR #330, opened 2026-04-23
Fix dual-source timestamp inflation in meeting mode

materemias:
## Summary
- Per-chunk start offset was derived from `transcript.duration_ms()`, which returns the max `end_ms` across **all** segments — conflating mic and loopback on dual-track captures. Every new chunk got pushed past the other source's end, roughly doubling apparent meeting length (observed: 80-minute meeting exporting as 2h17m).
- Track the offset **per `AudioSource`** so mic and loopback each advance on their own wall-clock timeline, by the actual audio duration consumed each chunk.
- Pad loopback chunks in `daemon.rs` to the mic chunk length with silence before dispatch — keeps both offsets advancing in lockstep if the loopback monitor lags at startup or briefly gaps. VAD still skips transcription on silent frames.

## Why
Repro: run `voxtype meeting export latest --format markdown --timestamps` on any meeting where both sides speak. Real-time captures with ~1h20m of audio were exporting transcript timestamps up to `[02:17:29]`.

## Test plan
- [x] `cargo build --lib`
- [x] `cargo test --lib meeting::` — 175 passing
- [x] `cargo clippy --lib --no-deps` — no new warnings
- [ ] Record a dual-source meeting, export, verify timestamps track real wall clock

Co-authored-by: Codex <noreply@openai.com>

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -2688,9 +2688,16 @@ impl Daemon {
                         if self.meeting_mic_buffer.len() >= chunk_samples {
                             let mic_chunk: Vec<f32> = self.meeting_mic_buffer.drain(..chunk_samples).collect();
 
-                            // Also drain loopback buffer up to the same amount
+                            // Also drain loopback buffer up to the same amount.
+                            // Pad with silence so it matches the mic chunk length —
+                            // keeps the per-source timestamp offsets advancing in
+                            // lockstep with wall-clock even if the loopback monitor
+                            // has a short startup lag or a transient gap.
                             let loopback_len = self.meeting_loopback_buffer.len().min(chunk_samples);
-                            let loopback_chunk: Vec<f32> = self.meeting_loopback_buffer.drain(..loopback_len).collect();
+                            let mut loopback_chunk: Vec<f32> = self.meeting_loopback_buffer.drain(..loopback_len).collect();
+                            if loopback_chunk.len() < chunk_samples {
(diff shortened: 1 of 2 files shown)


materemias:
Addressed Copilot review in 431a5f5.

**Fixed (P1)** — `src/daemon.rs` padding dropped. Padding a short loopback slice (e.g. 2s) to 30s of silence would dilute the VAD speech-frame ratio below the 10% threshold and cause real remote audio to be discarded. Now we drain only actual samples and reconcile timestamps separately.

**Fixed (P2)** — `src/meeting/mod.rs` offsets are now advanced up front, from the input sample count, *before* any fallible processing. If transcription errors, the caller has still drained the samples, and the offset now correctly reflects the elapsed time.

**Added** — `MeetingDaemon::sync_source_offsets()` levels all per-source offsets to the furthest-advanced one. Called after each mic+loopback iteration in the daemon loop, so a short or skipped loopback chunk catches up to wall-clock before the next chunk dispatches. Replaces the padding approach while solving the same "loopback monitor lags at startup" edge case.

Not addressed — the unit-test suggestion. Adding a focused test for this path would require building a mock `Transcriber` harness just for one test; the cost/value didn't pencil out given the existing integration coverage. Happy to add one if you'd prefer.

peteonrails:
This initially looks great. I need a few days to give it a closer look and test. THANK YOU

peteonrails (approved):

PR #341, opened 2026-05-01
Fix meeting audio capture and source diarization

sjug:
## Description

This PR fixes several meeting-mode issues around mic capture, final chunk handling, voice activity detection tuning, and source-based diarization.

The changes are split into four commits:

1. Honor `[meeting.audio].mic_device`
2. Flush buffered meeting audio on stop
3. Make meeting VAD threshold configurable
4. Preserve simple diarizer segment boundaries

### Honor `[meeting.audio].mic_device`

`mic_device` already existed in config and docs, but meeting mode always passed the main `[audio]` config into `DualCapture`. That meant a meeting-specific mic device override was ignored.

This now builds the meeting capture config at the call site:

- `"default"` and `""` both mean “use the normal dictation audio device”
- explicit values override the mic used for meeting capture only
- an INFO log is emitted only when the override is active, showing both the meeting mic and dictation mic

This keeps the normal/default path unchanged.

### Flush buffered meeting audio on stop

`DualCapture::stop()` returns any samples that arrived since the last poll, but `stop_meeting()` previously discarded them. That could drop speech near the end of a meeting, especially for short recordings or when the user stops shortly after speaking.

This change:

- keeps samples returned by `capture.stop().await`
- appends them to the meeting buffers
- flushes the final partial chunk before saving/stopping the meeting
- extracts shared chunk processing into helpers so the periodic poll path and final flush path use the same logic

### Make meeting VAD threshold configurable

Meeting chunks use a simple RMS gate before transcription. The existing default remains `0.01` for backwards compatibility, but quiet microphones can now tune it via:

```toml
[meeting.audio]
vad_threshold = 0.001
```

The threshold is wired through:

- user config: `MeetingAudioConfig`
- daemon meeting config construction
- CLI meeting command config construction
- internal `MeetingConfig`
- `ChunkConfig`

This also improves diagnostic logging for skipped meeting chunks by including source, duration, RMS, and threshold at debug level. Transcription logs now include the audio source (`Microphone` or `Loopback`) so mic/loopback behavior is easier to debug.

### Preserve simple diarizer segment boundaries

The simple diarizer assigns speakers based on audio source:

- microphone -> `You`
- loopback -> `Remote`

It previously merged adjacent same-speaker transcript segments. However, the caller applies diarization results back to transcript segments positionally with `zip()`. If the diarizer returned fewer segments than ASR produced, later transcript segments silently missed `speaker_id`.

This change removes merging from `SimpleDiarizer`, preserving a 1:1 mapping between transcript segments and diarized segments. As a result, every mic/loopback segment receives the expected speaker label.

## Related Issue

N/A

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update

## Testing

Commands run:

```bash
cargo check
cargo test meeting::chunk::
cargo test meeting::diarization::simple::
cargo test config::tests::test_meeting
cargo test config::tests::test_parse_meeting_config_with_nested_sections
cargo test
```

Results:

- `cargo check` passed.
- Targeted meeting/config tests passed.
- Full `cargo test` passed:
  - 539 unit tests passed
  - 25 integration tests passed
  - doc tests passed

Additional non-mutating checks run:

```bash
cargo fmt --check
cargo clippy --all-targets --all-features
```

Results:

- `cargo fmt --check` failed due existing unrelated formatting differences across the repo. This PR does not include those unrelated formatting changes.
- `cargo clippy --all-targets --all-features` failed because `--all-features` enables Whisper ROCm/HIP and this machine does not have `hipcc` in PATH. It also reported two existing `build.rs` `uninlined_format_args` warnings before failing.

Manual validation:

- Verified mic-only meeting recording produced transcript segments instead of zero segments.
- Verified mixed mic + loopback meeting recording produced both `You` and `Remote` speakers.
- Verified every exported transcript segment had a speaker label after the simple diarizer fix.

## Documentation

- [x] I have updated documentation as needed

Updated docs:

- `docs/CONFIGURATION.md`
- `docs/USER_MANUAL.md`
- `docs/MEETING_MODE.md`

## Additional Notes

The default meeting VAD threshold remains `0.01` to preserve existing behavior. Users with quiet meeting microphones can lower `[meeting.audio].vad_threshold`, with `0.001` documented as a suggested starting point.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -2780,11 +2780,22 @@ Echo cancellation mode for removing speaker bleed-through from the microphone si
 - `"auto"` - Use GTCRN neural speech enhancement on mic audio before transcription, followed by a phrase-level transcript dedup pass. The GTCRN model (~523 KB) is automatically downloaded on first `voxtype meeting start`.
 - `"disabled"` - No enhancement. Use this if you have system-level echo cancellation configured (e.g., PipeWire's `echo-cancel` module) or if you don't use loopback capture.
 
+### vad_threshold
+
+**Type:** Float
+**Default:** `0.01`
+**Required:** No
+
+RMS threshold for meeting chunk voice activity detection. Lower values are more permissive and can help quiet microphones; higher values skip more low-level noise before transcription. Set to `0.0` to disable this pre-transcription gate.
+
+For quiet USB/XLR mics, try `0.001`.
+
 **Example:**
 ```toml
 [meeting.audio]
 loopback_device = "auto"
 echo_cancel = "auto"  # GTCRN enhancement + transcript dedup
+vad_threshold = 0.001  # Optional: quiet mic tuning
 ```
 
 ---
--- a/docs/MEETING_MODE.md
+++ b/docs/MEETING_MODE.md
@@ -254,6 +254,10 @@ mic_device = "default"
 # Loopback device for capturing remote participants' audio
 # "auto" = auto-detect, "disabled" = mic only, or a specific device name
 loopback_device = "auto"
+
+# RMS threshold for meeting voice activity detection (default: 0.01)
+# Lower to 0.001 for quiet mics; set 0.0 to disable this pre-transcription gate
+vad_threshold = 0.01
 ```
 
 Setting `loopback_device = "auto"` lets voxtype capture system audio (the other side of a call). When loopback is active, speaker attribution can distinguish between "You" (from the mic) and "Remote" (from system audio).
--- a/docs/USER_MANUAL.md
+++ b/docs/USER_MANUAL.md
@@ -2205,6 +2205,7 @@ max_duration_mins = 180          # Maximum meeting length (0 = unlimited)
 mic_device = "default"           # Microphone (uses audio.device if not set)
 loopback_device = "auto"         # Capture remote participants: "auto", "disabled", or device name
 echo_cancel = "auto"             # GTCRN neural enhancement + transcript dedup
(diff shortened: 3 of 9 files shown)


peteonrails:
This is really good work that I may not get a chance to test before 0.7.0 drops, but I will try to get it in there.

peteonrails:
Rebased this onto current \`dev\` for you — the branch was 317 commits behind. One conflict in \`src/daemon.rs\` from #330 (dual-source timestamp inflation fix, merged this week) refactoring the same meeting-mode chunk-processing block you refactored into helpers here. Resolved by keeping your \`process_meeting_audio_pair\` / \`process_buffered_meeting_audio\` helper structure and folding in the \`sync_source_offsets()\` call #330 added — that call now lives inside \`process_meeting_audio_pair\` after both chunks are processed and before bleed-through dedup. Annotated with a comment pointing at #330.

The other three commits (\`Honor meeting audio mic device\`, \`Make meeting VAD threshold configurable\`, \`Preserve simple diarizer segment boundaries\`) applied cleanly with no conflicts.

Local gates: \`cargo fmt --check\` clean, \`cargo clippy --all-targets --no-deps -- -D warnings\` clean, \`cargo test --lib\` passes 717/718 with only the pre-existing environmental \`setup::parakeet::tests::detect_available_backends_returns_vec\` failure (unrelated, exists on \`dev\` too).

Your four-commit history is preserved (author still you on each commit). CI should kick off shortly — happy to merge once it's green. Thanks for the patient wait!

Issue #352, opened 2026-05-10
0.7.0 vulkan build breaks on Nix: glib-sys / gdk-pixbuf-sys pulled in unconditionally


bashfulrobot:
Voxtype 0.7.0's vulkan package fails to build on NixOS. 0.6.6 built fine on the same workstation.

Reproduce:

```
nix build github:peteonrails/voxtype/v0.7.0#packages.x86_64-linux.vulkan
```

The cargo build fails when compiling `glib-sys`. The crate's build script calls pkg-config looking for `glib-2.0`, pkg-config doesn't find `glib-2.0.pc`, the script exits 1.

Excerpt from the cargo log:

```
> pkg-config exited with status code 1
>   > PKG_CONFIG_PATH=/nix/store/...alsa-lib-1.2.15.3-dev/lib/pkgconfig:/nix/store/...openssl-3.6.1-dev/lib/pkgconfig:/nix/store/...vulkan-loader-1.4.341.0-dev/lib/pkgconfig PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags glib-2.0 'glib-2.0 >= 2.56'
>
>   The system library `glib-2.0` required by crate `glib-sys` was not found.
>   The file `glib-2.0.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
```

`PKG_CONFIG_PATH` at the failure point contains only the inputs the vulkan derivation declares in `flake.nix`: alsa-lib, openssl, vulkan-loader.

I tried adding `glib` and `pkg-config` to the vulkan derivation's `nativeBuildInputs` and `buildInputs`. With that change the build progresses past `glib-sys` but then fails the same way on `gdk-pixbuf-sys` looking for `gdk-pixbuf-2.0.pc`.

Workaround: pinned to adf0ea6 (pre-0.7.0). Happy to test a fix branch.

peteonrails:
Thanks @bashfulrobot - looking in to it now

peteonrails:
Diagnosed and fixed in https://github.com/peteonrails/voxtype/commit/39015cd, merged to main as bacc63a.

[...]

PR #353, opened 2026-05-11
flake: add missing deps

andresilva:
Add missing dependencies for the flake builds.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,7 +31,15 @@
 
         commonBuildInputs = with pkgs; [
           alsa-lib
+          atk
+          gdk-pixbuf
+          glib
+          gtk3
+          libx11
+          libxi
+          libxtst
           openssl
+          pango
         ];
 
         # Runtime dependencies wrapped into PATH


peteonrails:
@andresilva Thanks for this! A few of these dependencies were introduced by mistake, and I have gated them on the macos feature flag here: https://github.com/peteonrails/voxtype/commit/bacc63aaba897dbe3a7e758c890a9d30e4ff2e09

peteonrails:
Obsoleted by the v0.7.1 fix landed in 39015cd / bacc63a. That commit moved `tray-icon` and `rdev` to `[target.'cfg(target_os = "macos")'.dependencies]` — those crates were the reason `glib-sys`, `gdk-pixbuf-sys`, `atk-sys`, `gtk3-sys`, `libx11`, `libxi`, and `libxtst` were getting pulled into every Linux build. With them gated out, none of the GTK3 / X11 system libs need to be in `commonBuildInputs` anymore.

Closing this since the deps you added are all in that obsolete set. See #352 for the full root-cause writeup.

The OSD packages from your follow-up PR and the GTK4 visibility fix were landed directly on main (2312533 and a06f82d respectively), with your authorship preserved. Thanks for the clean stacked commits — they were trivial to cherry-pick.

andresilva:
Makes sense, I was a bit lazy to debug this, it also didn't make much sense to me that I was requiring X11 deps to use this on wayland.

PR #354, opened 2026-05-11
flake: package osd binaries

andresilva:
Package the OSD binaries.

Merge after #353.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,7 +31,15 @@
 
         commonBuildInputs = with pkgs; [
           alsa-lib
+          atk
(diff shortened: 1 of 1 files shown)


peteonrails:
Landed as 2312533 on main with your authorship preserved. Cherry-picked just the `fe331db` commit on top of the v0.7.1 fix so the deps from the parent PR were not pulled in. The osdGtk4 derivation builds fine without them because `gtk4-layer-shell` propagates the gtk4 → glib/pango/gdk-pixbuf chain transitively. Thanks!

PR #355, opened 2026-05-11
Fix GTK4 OSD staying visible on startup until first use

andresilva:
`window.present()` maps the surface at startup, but the `visible` Cell tracking GTK visibility was initialized to `false`. On the first idle tick, the hide branch checks `if visible.get()` before calling `set_visible(false)`, so it skipped the hide and the window stayed mapped until the first frame arrived and synced the flag.

Merge after #354.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,7 +31,15 @@
 
         commonBuildInputs = with pkgs; [
           alsa-lib
+          atk
+          gdk-pixbuf
+          glib
+          gtk3
+          libx11
+          libxi
+          libxtst
           openssl
+          pango
         ];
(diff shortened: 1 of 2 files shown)


peteonrails:
Landed as a06f82d on main with your authorship preserved. Cherry-picked just `8b36ff2` on top of v0.7.1, dropping the parent deps commit. Thanks for tracking down the OSD startup-visibility bug.

andresilva:
v0.7.1 doesn't include this or #354, it's tagged before those two commits. Looking at the release notes I think that was an oversight.

PR #379, opened 2026-05-17
Investigation: Quickshell engine picker blocked on missing config-set CLI

peteonrails:
## Summary

Investigation report (no shippable widget). Engine switching has no
programmatic CLI/IPC entry point that a Quickshell widget can invoke,
so the v0.7.3 Wave 2 engine picker can't be built without prior work.

The full reasoning, rejected alternatives, and recommended two-PR
sequence are in `docs/investigations/quickshell-engine-picker.md`.

### What I found

- `--engine` flag and `VOXTYPE_ENGINE` env var are per-invocation
  overrides, not persistent.
- The persistent path is `engine = "..."` in `~/.config/voxtype/config.toml`.
- The TUI's `src/tui/engine.rs` writes that field via `ConfigEditor`
  (atomic temp+rename via `toml_edit`), but `ConfigEditor` has no CLI
  surface — it's `pub(crate)` and only invoked from the interactive TUI.
- `src/daemon.rs` handles SIGUSR1/SIGUSR2/SIGTERM only. No SIGHUP, no
  config-mutation Unix socket. The daemon must be restarted to pick up
  a new engine.

### Why I didn't ship a half-baked widget

Pete's brief said: "If you find that engine switching genuinely doesn't
have a CLI/IPC entry point yet, STOP and write a short investigation
report... and PR that report as a draft rather than ship a half-baked
widget."

The only ways to ship the picker today are:

1. Have QML's `Process` shell out to `sed`/`python3` to patch the TOML.
   No validation, no feature gating, breaks on minimal Hyprland setups
   without Python, mangles comments/ordering. Half-baked.
2. Have QML's `Process` spawn `voxtype configure` and somehow drive the
   TUI to the engine pane. Brittle and worse UX than just opening the
   TUI manually.

### Recommended path forward

1. Land `voxtype config set engine <name>` as a small Rust-only PR
   (~50 LOC). Promotes `ConfigEditor` to a public CLI surface, reuses
   the TUI's feature-gating logic to reject engines the binary wasn't
   built with.
2. Then the Quickshell picker is trivial: `Process { command: ["voxtype",
   "config", "set", "engine", selected, "--restart"] }` + a `Process`
   that reads `voxtype info variants --json` (or a new `voxtype info
   engine --json`) to populate the list.

Refs `project_quickshell_v073.md` item 2.

## Test plan
- [ ] Pete reviews the recommendation and decides between: (a) land
      `voxtype config set` in v0.7.3, (b) defer picker to v0.7.4, or
      (c) ship a placeholder picker that opens `voxtype configure`.
- [ ] If (a): open follow-up issue for the CLI work.
- [ ] If (b)/(c): merge this investigation as the record of why the
      picker isn't in Wave 2.

## Open question

Is there value in adding a non-systemd restart path? On macOS,
non-systemd Linux setups (init scripts, runit, OpenRC, plain
`voxtype daemon` in tmux), and inside containers, `systemctl --user
restart voxtype` isn't the right verb. A `--reload` flag that sends
a signal the daemon hasn't yet defined (SIGHUP → "re-exec myself")
would be more portable, but that's its own design conversation.

### DIFF
--- a/docs/investigations/quickshell-engine-picker.md
+++ b/docs/investigations/quickshell-engine-picker.md
@@ -1,0 +1,190 @@
+# Investigation: Quickshell engine picker — missing CLI/IPC entry point
+
+**Status:** Blocked. Engine switching has no programmatic CLI/IPC entry point
+that the Quickshell engine picker can invoke. Recommend adding a small
+`voxtype config set` CLI surface before shipping the widget.
+
+**Wave:** v0.7.3 Wave 2 (Quickshell composition pass)
+**Related:** `project_quickshell_v073.md` item 2
+
+## Goal
+
+Wave 2 of the Quickshell v0.7.3 work ships an engine picker — a QML panel
+that lists installed transcription engines (whisper, parakeet, moonshine,
+sensevoice, paraformer, dolphin, omnilingual, cohere), marks the active
+one, and lets the user click another engine to switch.
+
+The QML side is straightforward (`PanelWindow` + `ListView` + a `Process`
+to invoke whatever switches the engine). What's missing is the thing the
+`Process` is supposed to invoke.
+
+## What exists today
+
+### 1. CLI flags (per-invocation only)
+
+`src/cli.rs` exposes a top-level `--engine` flag and a per-subcommand
+`--engine` flag on `voxtype transcribe`:
+
+```rust
+/// Override transcription engine
+#[arg(long, value_name = "ENGINE")]
+pub engine: Option<String>,
+```
+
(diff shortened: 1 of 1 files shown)


peteonrails:
Superseded by #385 which lands EnginePicker.qml. The CLI (#382) and compiled_features fix (#384) it depended on are in dev. Safe to close.

peteonrails:
Superseded by #385 which shipped the engine picker on top of #382's `voxtype config set engine` CLI. Investigation report on the original branch was the right call at the time; closing as resolved.

PR #410, opened 2026-05-19
output/dotool: route through dotoolc when dotoold is running

materemias:
## Summary

When `dotoold` is running, voxtype now routes text injection through `dotoolc` instead of spawning fresh `dotool` processes per call. Each direct `dotool` invocation pays ~700ms of uinput keyboard setup; for streaming backends (Parakeet, upcoming Soniox) that fire 60+ output() calls per session, this stacked to 40+ seconds of typing latency. Via `dotoolc` the setup cost is paid once at daemon startup and per-call latency drops to sub-10ms.

## How it works

- `DotoolOutput::daemon_pipe_path()` detects a live daemon: the FIFO must exist AND `O_WRONLY|O_NONBLOCK` open must succeed. A crashed `dotoold` leaves the FIFO on disk; the kernel returns ENXIO on a non-blocking write-open in that case, so we cleanly fall back to direct `dotool`.
- `live_daemon_pipe_path()` exposes the same check to the cancel-rewind path in `output/streaming.rs::try_dotool_backspaces`. On KDE Plasma, raw `dotool` creating ephemeral uinput keyboards on every backspace burst can cause the compositor to drop events; routing through the persistent dotoold keyboard avoids that.
- Users without `dotoold` see no behavior change — direct `dotool` fallback is untouched.

## Drive-by

`tracing::info!` on a successful type now reports `text.chars().count()` instead of `text.len()` so multi-byte Unicode (Hungarian, CJK, …) gets the right character count.

## Test plan

- [ ] `cargo test --lib output::dotool` passes
- [ ] Without \`dotoold\` running: dictation still works via fallback \`dotool\` invocation
- [ ] With \`dotoold\` running: log line shows \`Text typed via dotoolc (N chars)\`; uinput keyboards listed via \`libinput list-devices\` stay at the persistent daemon's keyboard
- [ ] Long streaming session (Parakeet streaming, 5 min) doesn't accumulate ~700ms per-call latency
- [ ] Cancel mid-stream backspaces visible chars cleanly via dotoolc

## Notes

- Content-only on \`src/output/dotool.rs\` + a small hunk in \`src/output/streaming.rs::try_dotool_backspaces\`. No config changes.
- Local builds on rustc 1.91 hit a pre-existing \`__cpuid\` unsafe-block error in \`src/cpu.rs\` introduced by \`871ee5c\` — orthogonal to this PR; worth fixing separately.

### DIFF
--- a/src/output/dotool.rs
+++ b/src/output/dotool.rs
@@ -42,6 +42,18 @@ use std::time::Duration;
 use tokio::io::AsyncWriteExt;
 use tokio::process::Command;
 
+/// Truncate a string for log emission, replacing newlines with literal `\n`
+/// so multi-line dotool command streams stay on one log line. Keeps the
+/// first `max_chars` Unicode scalars and appends an ellipsis when cut.
+fn truncate_for_log(s: &str, max_chars: usize) -> String {
+    let one_line = s.replace('\n', "\\n");
+    if one_line.chars().count() <= max_chars {
+        return one_line;
+    }
+    let head: String = one_line.chars().take(max_chars).collect();
+    format!("{}…", head)
+}
+
 /// dotool-based text output with keyboard layout support.
 pub struct DotoolOutput {
     /// Delay between keypresses in milliseconds
@@ -157,7 +169,16 @@ impl TextOutput for DotoolOutput {
         let pipe = Self::daemon_pipe_path();
         let binary = if pipe.is_some() { "dotoolc" } else { "dotool" };
         let set_layout_env = pipe.is_none();
-        tracing::debug!(target: "voxtype::dotool::wire", "-> {:?}", commands);
+        // Wire trace only at the TRACE level so the user's typed text
+        // isn't dumped to logs on the default -vv (DEBUG) verbosity.
+        // The dotool command stream contains every typed character.
+        if tracing::enabled!(target: "voxtype::dotool::wire", tracing::Level::TRACE) {
+            tracing::trace!(
+                target: "voxtype::dotool::wire",
+                "-> {:?}",
+                truncate_for_log(&commands, 40)
+            );
+        }
 
         let mut cmd = Command::new(binary);
         cmd.stdin(Stdio::piped())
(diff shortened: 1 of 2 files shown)


materemias:
Pushed a fixup commit (`4cb5b90`) addressing both Copilot suggestions:

**1. Wire-trace privacy leak (`src/output/dotool.rs:160`)**

The `debug!(target: "voxtype::dotool::wire", ...)` was dumping the full dotool command stream — including every typed character — under the default `-vv` verbosity. That's a real privacy regression compared to the prior 20-char preview.

Mitigations stacked:

- Downgraded the log to `trace!`, so it now only fires on `-vvv`.
- Added a `truncate_for_log(s, max_chars)` helper that caps the logged payload at 40 scalars and escapes newlines so the wire trace stays a single line.
- Gated the log with `tracing::enabled!` so the helper isn't called unless the target is actually live at TRACE.

End result: default `-vv` no longer leaks dictation text; `-vvv` shows the wire path with bounded payload.

**2. Parallel-test env-var race (`src/output/dotool.rs:291`)**

The `daemon_pipe_detection_respects_env_var` test was mutating process-wide `DOTOOL_PIPE` without serialisation. No other test reads `DOTOOL_PIPE` today, but the pattern would flake the moment a second test touches the same var.

Fix:

- Module-local `static DOTOOL_PIPE_LOCK: Mutex<()>` serialises any test that mutates this env var.
- New `DotoolPipeEnvGuard` RAII type sets the value on construction and restores the prior value (or unsets) on drop — so a panicking test can't pollute the run.
- Poisoned-lock case is handled with `unwrap_or_else(|e| e.into_inner())`, so one panic doesn't break every subsequent test in the file.

The test itself collapses to:

\`\`\`rust
let _guard = DotoolPipeEnvGuard::set("/nonexistent/dotool-pipe-test");
assert!(DotoolOutput::daemon_pipe_path().is_none());
\`\`\`

All 7 CI checks (fmt, clippy, test, build cpu/onnx, ci-success, nix-success) are green on `4cb5b90`. PR is ready for human review.

peteonrails:
When I was testing streaming with `dotool` it was pretty unbearable: I am glad you got this figured out! it's one more step toward leaving the "experimental" phase with streaming. 

Merging to dev and will plan to include it in 0.7.4

PR #411, opened 2026-05-19
Add Soniox cloud streaming STT backend

materemias:
## Background

Picks up the unanswered question from #326 ("Staying local only or expanding
to cloud providers"). This PR is a concrete proposal for one half of the
answer: add a cloud STT provider as a configurable engine choice, not as a
replacement for local models.

Voxtype's local backends do solid work in English. For non-English languages
(Hungarian is my own daily use, which is why I opened #326), Whisper's raw
output isn't quite paste-ready: word boundaries, casing, punctuation, and
proper nouns all need cleanup. The existing workaround is the `post_process`
command hook driving an LLM, typically Gemini Flash or similar.

That path works but has two real costs:

1. **Latency.** On my local hardware (Vulkan), Whisper v3 large turbo takes
   2-3 s for the actual transcription, and Gemini Flash post-processing adds
   another 1.5-2 s. End-to-end that's 4-5 s of perceived wait per utterance
   before paste-ready text reaches the cursor.
2. **Cost.** Every chunk pays the LLM provider's input + output tokens on
   already-transcribed text. For long meetings this adds up fast.

A streaming-quality cloud STT provider that produces paste-ready non-English
output without an LLM post-processing stage solves both: lower end-to-end
latency, lower cost per minute, no separate post-process pipeline to maintain.

[v0.7.2](https://github.com/peteonrails/voxtype/releases/tag/v0.7.2) made
this practical. The streaming dictation release shipped the
`StreamingTranscriber` trait, the `StreamingEvent::{Partial, Final, Ended,
Error}` channel, `StreamingSession` for delta typing + cancel-rewind, and the
PTT-to-toggle auto-promote gate, all aimed at Parakeet streaming (which is
English-only in practice). Once that infrastructure was in place, plugging a
cloud streaming provider in as a second consumer of the same plumbing became
straightforward: implement `StreamingTranscriber`, emit the same events,
inherit all the existing daemon and OSD wiring for free.

Soniox fits the criteria. Realtime WebSocket with sub-second partial latency,
60+ languages with native casing/punctuation, per-token finality flags so the
local reconciler can do live partial typing, and built-in vocabulary boosting
that replaces much of what the post-processing prompt was doing.

## What this PR adds

Two Soniox APIs, two voxtype use cases:

- **Dictation** uses the realtime WebSocket API (`stt-rt-v4`). Live partial
  typing at the cursor, sub-second latency, toggle activation.
- **Meetings** use the async REST API (`stt-async-v4`) automatically. Full
  chunk context gives better diarization-friendly transcripts and
  audio-second billing. Users don't have to flip a flag, it's the natural
  fit.

\`engine = "soniox"\` becomes a first-class voxtype engine. Config surface
(defaults shown):

| Field | Default | Notes |
|---|---|---|
| \`api_key\` | env: \`SONIOX_API_KEY\` | Required |
| \`model\` | \`stt-rt-v4\` (or \`stt-async-v4\` when \`async_api\`) | Override only if pinning a build |
| \`language_hints\` | \`["hu", "en"]\` | Empty = full auto-detect |
| \`language_hints_strict\` | \`true\` | Bias output to hinted languages |
| \`streaming\` | \`true\` | Realtime live partials vs PTT-compatible one-shot |
| \`type_partials\` | \`true\` | Live cursor feedback during streaming |
| \`terms\` / \`terms_file\` | none | Vocabulary boost (proper names, jargon) |
| \`context\` | none | Free-form domain context |
| \`async_api\` | \`false\` | Use REST batch instead of WebSocket |
| \`async_max_wait_secs\` | \`120\` | Async job total timeout |

CLI flag \`--soniox-api-key\`, engine override \`--engine soniox\`.
Push-to-talk auto-promotes to toggle for streaming mode (matches existing
parakeet-streaming behavior); set \`[soniox] streaming = false\` to keep PTT.

### Meeting mode auto-switch

\`Config::with_meeting_mode_overrides()\` returns a clone of the config with
engine-specific tweaks for long-form transcription. For Soniox it forces
\`async_api = true\` regardless of the user's setting. No config flag for
"meeting uses realtime" because opening a fresh WS session per 30 s chunk
has no upside over async.

### Streaming-output extension

The \`StreamingEvent\` enum gains a \`Replace { backspace, text }\` variant.
Soniox occasionally finalizes a non-final tail with revised punctuation or
word choice; \`Replace\` lets the streaming output session emit backspaces
and re-type cleanly without losing cursor state. Parakeet streaming, which
never revises, ignores the new variant.

\`Config::streaming_active()\` is generalized from a hardcoded
\`parakeet.streaming\` check to a per-engine match. Future streaming backends
plug into the PTT auto-promote gate without editing the daemon.

### OSD visibility during async drain

When the user releases the hotkey on a streaming Soniox session, the mic is
cut immediately but the server may still be emitting trailing finals for a
second or two. A small "silence pump" task publishes silent OSD frames during
that drain window so the on-screen display stays visible until the finalized
text actually arrives at the cursor. Without this, the OSD would fade out
as soon as the audio stream stopped and the user wouldn't know to wait.

## Architecture notes

- \`src/transcribe/soniox.rs\` (~1500 lines, new) implements both \`Transcriber\`
  (sync trait) and \`StreamingTranscriber\`. The sync-to-async bridge uses
  \`tokio::runtime::Handle::try_current()\` + \`block_in_place\` to reuse the
  meeting daemon's ambient runtime, falling back to a private current-thread
  runtime for the one-shot CLI path.
- Token reconciliation: Soniox sends finals once (non-cumulative) and
  revisable non-finals. The reconciler tracks only \`typed_partial\` (the
  in-flight non-final tail) and emits \`Partial\` / \`Final\` / \`Replace\`
  based on the LCP between server state and what's already at the cursor.
- Cleanup: \`DELETE /v1/transcriptions/{id}\` cascades to the file, so we
  issue only the transcription delete. No separate \`/files/{id}\` delete
  unless the transcription was never created (orphan upload case).
- Wire-trace log redacts \`api_key\` via a serde_json round-trip.

## Tests

41 unit tests in \`src/transcribe/soniox.rs\` cover the reconciler state
machine, init-frame shape including \`language_hints_strict\` and
\`terms_file\` plumbing, async/realtime mode switching including the meeting
auto-switch, special-token filtering (\`<end>\`, \`<fin>\`, \`<lang:xx>\`,
\`<speaker:N>\`), and WAV / f32-to-s16le encoding.

## Build features

New \`soniox\` cargo feature (default off) pulls \`tokio-tungstenite\` and
\`reqwest\` with rustls. TUI, menubar, and notification engine matches
updated for the new variant so macOS and Linux builds stay exhaustive.

## Docs

- New \`docs/SONIOX.md\` with config recipes, dictation/meeting routing table,
  and a note on dotoold's perf benefit for streaming output.
- \`docs/CONFIGURATION.md\` and \`docs/TROUBLESHOOTING.md\` extended with
  per-field reference and common failure modes (401, 429, etc.).
- \`docs/MODEL_SELECTION_GUIDE.md\` and \`docs/USER_MANUAL.md\` get a Soniox
  row in their engine tables.

## Scope

Standalone on \`dev\`. Does not depend on #410 (dotoolc fast path). When
#410 lands, Soniox streaming benefits from dotoolc routing for free via the
existing \`try_dotool_backspaces\` callsite, no Soniox-side changes required.

## Test plan

- [ ] \`cargo build --features soniox\` clean
- [ ] \`cargo test --features soniox --lib soniox\` passes 41/41
- [ ] \`SONIOX_API_KEY\` set, dictation with \`[soniox] streaming = true, language_hints = ["en"]\`: partial typing visible at cursor, finals stable
- [ ] \`[soniox] streaming = false\`: hotkey-release one-shot still works
- [ ] Bilingual dictation \`language_hints = ["hu", "en"]\`: mid-utterance language switches handled cleanly
- [ ] \`voxtype meeting start\` with \`engine = "soniox"\`: daemon log shows \`Soniox meeting mode: routing to async API (stt-async-v4); dictation path unchanged\`; async chunks complete and land in \`transcript.json\`
- [ ] Cancel mid-stream: typed partials backspaced cleanly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -273,6 +279,15 @@ name = "block"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
+
+[[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
 
 [[package]]
 name = "block2"
@@ -782,6 +797,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cpufeatures"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "crc32fast"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -854,6 +878,16 @@ name = "crunchy"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
+
+[[package]]
+name = "crypto-common"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78c8292055d1c1df0cce5d180393dc8cce0abec0a7102adb6c7b1eef6016d60a"
+dependencies = [
+ "generic-array",
+ "typenum",
+]
 
 [[package]]
 name = "cursor-icon"
@@ -946,6 +980,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c87e182de0887fd5361989c677c4e8f5000cd9491d6d563161a8f3a5519fc7f"
 
 [[package]]
+name = "data-encoding"
+version = "2.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4ae5f15dda3c708c0ade84bfee31ccab44a3da4f88015ed22f63732abe300c8"
+
+[[package]]
 name = "der"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -993,6 +1033,16 @@ checksum = "ab63b0e2bf4d5928aff72e83a7dace85d7bba5fe12dcc3c5a572d78caffd3f3c"
 dependencies = [
  "derive_builder_core",
  "syn 2.0.117",
+]
+
+[[package]]
+name = "digest"
+version = "0.10.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
+dependencies = [
+ "block-buffer",
+ "crypto-common",
 ]
 
 [[package]]
@@ -1492,6 +1542,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures-sink"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c39754e157331b013978ec91992bde1ac089843443c49cbc7f46150b0fad0893"
+
+[[package]]
 name = "futures-task"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1505,6 +1561,7 @@ checksum = "389ca41296e6190b48053de0321d02a77f32f8a5d2461dd38762c0593805c6d6"
 dependencies = [
  "futures-core",
  "futures-macro",
+ "futures-sink",
  "futures-task",
  "pin-project-lite",
  "slab",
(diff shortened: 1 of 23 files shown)


materemias:
Pushed two fixup commits since the initial PR:

**\`87b8188\` — fix clippy errors flagged by CI**

CI rejected on \`cargo clippy -- -D warnings\`. Fixes:

- \`config.rs\` tests for \`with_meeting_mode_overrides\`: rewrote with struct-init syntax instead of \`let mut cfg = Config::default(); cfg.field = ...\` so \`clippy::field_reassign_with_default\` is satisfied.
- \`soniox.rs\` hygiene cleanup (visible only under \`--features soniox\`, not CI's default-features clippy, but worth fixing anyway):
  - collapsed a nested \`if\` (\`clippy::collapsible_if\`)
  - dropped redundant \`&\` borrows on \`client: &reqwest::Client\` callsites (\`clippy::needless_borrow\`)
  - removed useless \`.into()\` on values already at the target type — \`Message::Text\` takes \`String\`, \`Message::Binary\` takes \`Vec<u8>\` (\`clippy::useless_conversion\`)

No behavior change.

**\`71aaeed\` — address Copilot review**

All 5 of Copilot's inline comments were valid:

1. **\`src/output/streaming.rs::replace_and_commit\`** — when \`emit_backspaces(n)\` returned 0 (no backspace-capable backend available), the code was still mutating \`self.typed_chars\` and truncating \`self.partial\`. That made the session's bookkeeping drift from the real cursor and broke subsequent cancel-rewind accounting. Now state is only mutated when backspaces actually emit; the no-backend case keeps state consistent with the visible cursor and just logs the warning.

2. **\`src/daemon.rs::start_streaming_drain_pump\` doc comment** — said the pump runs at 100 Hz, but the implementation in \`audio/levels.rs\` ticks at 30 Hz (changed during the /polish pass before initial submission). Updated comment to match.

3. **\`src/transcribe/soniox.rs\` module docs** — claimed the batch path ends with "empty binary end-of-audio marker". The actual implementation (and Soniox's protocol docs) ends with \`{"type":"finalize"}\` followed by an empty text frame. Doc rewritten to match.

4. **\`src/config.rs::streaming_active()\`** — Soniox arm was returning \`true\` when \`engine = "soniox"\` but \`[soniox]\` section missing (\`unwrap_or(true)\`). That would auto-promote PTT to toggle for a config that can't actually initialize the transcriber (\`create_transcriber\` would error on missing api_key). Now returns \`false\` when the section is missing, mirroring the Parakeet arm's explicit-opt-in shape.

5. **\`docs/TROUBLESHOOTING.md\`** — section on Soniox tail revision claimed \`eitype\` was backspace-capable (it isn't — \`output/streaming.rs::emit_backspaces\` only tries wtype/dotool/ydotool) and referenced a log line \`Soniox final … does not align with typed partial\` that doesn't exist in the codebase. Replaced with the actual fallback chain and the real warn message users should grep for if backspace fails.

Local validation across all features I touched:

- \`cargo fmt --check\` clean
- \`cargo clippy --all-targets --no-deps -- -D warnings\` clean (CI's command)
- \`cargo clippy --all-targets --no-deps --features soniox -- -D warnings\` clean (extra safety for feature-gated paths)
- \`cargo test --features soniox --lib soniox\` passes 41/41

CI re-running on \`71aaeed\`.

materemias:
note: Meeting mode diarization could work theoretically with Soniox, but only if you were one-shotting the full meeting with an asynchronous upload and transcription. Right now chunk-based meeting processing is the way to go with ML diarization (from my perspective, it needs some adjustments, so expect a related PR).

peteonrails (approved):

PR #418, opened 2026-05-20
meeting/diarization: VAD sub-windows + tunable ECAPA threshold

materemias:
## Summary

The ML speaker diarizer (`backend = "ml"`) was effectively unusable in real meetings: a 2 h 20 m capture earlier this month finished with 96% of segments labeled `Unknown` and four stray `SPEAKER_NN` labels that clustered by audio source rather than by voice. This PR fixes both root causes and exposes the relevant tuning knobs as config.

## Problem

Two compounding issues in `src/meeting/diarization/ml.rs`:

1. **30 s mega-segment feeding.** Each transcript segment from Soniox is roughly a full 30 s chunk's worth of audio. ECAPA-TDNN was trained on 2–5 s utterances. Averaging a 30 s mixed-speech blob into one embedding produces a noisy centroid that doesn't cluster against anything reliably.
2. **Hardcoded `similarity_threshold = 0.75`.** Combined with the noisy embeddings above, the cosine similarity between two same-speaker segments rarely cleared 0.75. `next_speaker_id` raced ahead until `max_speakers` was hit, then every subsequent embedding fell to `Unknown`.

The transcript looked plausible (Soniox kept transcribing fine) but diarization was effectively off.

## Changes

- New `super::vad_subwindows(samples, sample_rate, window_secs, hop_secs, rms_floor)` helper. RMS-gated overlapping windows. `hop_secs` is clamped to a minimum of 0.1 s so a misconfigured 0 cannot generate hundreds of thousands of windows per segment.
- `SpeakerEmbedding` gains a `count: u32` field and an online-mean centroid update on match. The previous code anchored each speaker on a single first-utterance embedding, which drifted on long recordings.
- `MlDiarizer::diarize` now splits each transcript segment into 4 s sub-windows (2 s hop by default), embeds each, votes by majority for the segment's speaker, hoists the state mutex once per segment, and breaks tied votes deterministically by lowest `SpeakerId::Auto(n)` (first-seen wins).
- `DiarizationConfig` and `[meeting.diarization]` gain four backward-compatible serde-defaulted knobs:

  | Field | Default | Purpose |
  |---|---|---|
  | `similarity_threshold` | `0.25` | cosine sim for matching new embeddings to existing speakers |
  | `vad_window_secs` | `4.0` | ECAPA sub-window length |
  | `vad_hop_secs` | `2.0` | sub-window hop |
  | `vad_rms_floor` | `0.005` | silence gate |

  All four are no-ops when `backend != "ml"`.

## Validation

Empirical sweep against five YouTube clips covering different speaker counts and conditions:

| Source | Duration | Expected | Detected @ default 0.25 |
|---|---|---|---|
| EN 4-person roundtable | 2:00 | 4 | **4** |
| HU 3-person | 2:00 | 3 | **3** (1 single-window blip) |
| 1 h talk | 58:05 | unknown | **5** (4 stable + 1 noise) |
| EN 4-person podcast | 22:41 | 4 | 3 at 0.25; **4** at 0.30 |
| 10-person Google Meet | 1:17:21 | 10 | **10** with plausible time distribution |

For studio recordings with similar voices, raising `similarity_threshold` to 0.30–0.35 hits the right count. The defaults are conservative enough for noisier real-meeting audio.

The 17 unit tests in `meeting::diarization` pass. `cargo build`, `cargo test`, `cargo fmt`, and `cargo clippy` all clean (no new warnings introduced by this PR; three pre-existing warnings remain).

## Known limitations (not addressed here)

- `MlDiarizer` ignores the `_source` argument, so mic and loopback share the same speaker state. If you and a remote sound alike, they can merge at threshold 0.25. Raising the threshold or using `backend = "simple"` are the workarounds.
- First-chunk-wins ordering means `SPEAKER_00` is whoever spoke first; identities aren't stable across meetings.
- Codec asymmetry (Opus-encoded loopback vs raw mic) can shift embeddings slightly even for the same voice.

Follow-up ideas: hybrid backend (source-keyed `You`/`Remote` plus ECAPA only on loopback for multi-remote), pre-enrolled speaker bank, post-meeting agglomerative re-clustering, Silero VAD instead of energy gating, larger embedding model.

## Build note

This PR is independent of PR #341, but the current `dev` tip fails to build on rustc 1.91+ because `src/cpu.rs` calls `std::arch::x86_64::__cpuid` without an `unsafe` block. That's a separate one-line fix not included here. CI on older toolchains should be unaffected.

## Test plan

- [ ] `cargo build --release --features ml-diarization,onnx-load-dynamic` on a host with `onnxruntime-cpu` installed
- [ ] `cargo test --features ml-diarization,onnx-load-dynamic --lib meeting::diarization`
- [ ] Run a 1-on-1 meeting with `backend = "ml"` and verify two clean clusters (You + Remote)
- [ ] Run a multi-person meeting and verify the right number of distinct labels appear
- [ ] Verify users on `backend = "simple"` or `"remote"` see no behavior change (the new knobs are silently ignored)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1697,6 +1697,24 @@ pub struct MeetingDiarizationConfig {
     /// Minimum segment duration in milliseconds for ML embedding extraction
     #[serde(default = "default_min_segment_ms")]
     pub min_segment_ms: u64,
+
+    // The four fields below apply only to backend = "ml"; the "simple" and
+    // "remote" backends ignore them.
+    /// Cosine similarity threshold for the ML backend (0.20-0.30 typical for ECAPA on 4s windows)
+    #[serde(default = "default_similarity_threshold")]
+    pub similarity_threshold: f32,
+
+    /// VAD sub-window length in seconds for ECAPA feeding
+    #[serde(default = "default_vad_window_secs")]
+    pub vad_window_secs: f32,
+
+    /// VAD sub-window hop in seconds
+    #[serde(default = "default_vad_hop_secs")]
+    pub vad_hop_secs: f32,
+
+    /// RMS floor for treating a sub-window as silence
+    #[serde(default = "default_vad_rms_floor")]
+    pub vad_rms_floor: f32,
 }
 
 fn default_diarization_backend() -> String {
@@ -1709,6 +1727,25 @@ fn default_max_speakers() -> u32 {
 
 fn default_min_segment_ms() -> u64 {
     500
+}
+
+// Empirically tuned against multi-speaker test clips (4-person roundtable,
+// 3-person panel, 1h talk). The previous 0.75 anchor was far too strict for
+// 4s ECAPA windows and produced overwhelmingly Unknown labels in practice.
+fn default_similarity_threshold() -> f32 {
+    0.25
+}
+
+fn default_vad_window_secs() -> f32 {
+    4.0
+}
+
+fn default_vad_hop_secs() -> f32 {
+    2.0
+}
+
+fn default_vad_rms_floor() -> f32 {
+    0.005
 }
 
 fn default_chunk_duration() -> u32 {
@@ -1731,6 +1768,10 @@ impl Default for MeetingDiarizationConfig {
             max_speakers: default_max_speakers(),
             model_path: None,
             min_segment_ms: default_min_segment_ms(),
+            similarity_threshold: default_similarity_threshold(),
+            vad_window_secs: default_vad_window_secs(),
+            vad_hop_secs: default_vad_hop_secs(),
+            vad_rms_floor: default_vad_rms_floor(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


materemias:
Addressed both review notes in 7fedba3:

**1. Mutex critical section narrowed.** ECAPA inference is the slow step (~150 ms × ~14 sub-windows per segment); holding the speaker-state lock across all of those calls was the wrong shape. The loop now runs in two passes — first gather all sub-window embeddings without touching the state lock, then take the lock once and process the cluster updates sequentially.

Sequential processing inside the lock is preserved on purpose: the online centroid mean update has to observe earlier matches within the same segment, so each `find_or_create_speaker` call needs to see the centroid mutations from the prior call. Parallelizing them would silently change the clustering behavior.

**2. Poison error context restored.** The warn log now includes the `PoisonError` via `{}` formatting again, so a future poisoning has the underlying panic message in the journal instead of just a bare "lock poisoned" line.

Tests in `meeting::diarization` still pass (17/17), no behavior change in the steady-state path — only contention shape differs.

PR #420, opened 2026-05-20
Add --diarization flag to meeting start for per-meeting backend switch

materemias:
## Summary

- Add `--diarization` flag (and `VOXTYPE_MEETING_DIARIZATION` env var) to `voxtype meeting start`, accepting `simple` or `ml`
- CLI writes a sibling trigger file `meeting_start_diarization` *before* the existing `meeting_start` trigger; the daemon reads both atomically and overrides `[meeting.diarization].backend` for that meeting only
- Refactor `check_meeting_start` to return a `MeetingStartTrigger { title, diarization }`; extract a small `read_trimmed_nonempty` helper used by both fields

## Why

The diarization backend was only configurable via the config file, which made the choice static. In real use, the right backend depends on the meeting:

- **1:1 calls** — ML diarization is overkill and can over-fragment the transcript across "speakers" that are really just the same person at different acoustic distances from the mic. Source-based attribution (`simple` → You vs Remote) is the safer baseline.
- **Multi-person meetings** — `ml` actually earns its complexity.

Real usage flips between these case by case, so the choice belongs on the CLI, not in a config file.

## Usage

\`\`\`bash
voxtype meeting start --diarization simple        # 1:1, You vs Remote
voxtype meeting start --diarization ml --title "team standup"
VOXTYPE_MEETING_DIARIZATION=ml voxtype meeting start
\`\`\`

Omitting the flag falls back to \`[meeting.diarization].backend\` in config — existing setups are unchanged.

## Implementation notes

- Validation lives in clap (\`value_parser = [\"simple\", \"ml\"]\`); the daemon trusts the file contents on read.
- Write order is load-bearing: diarization sibling file is written first so the daemon never observes \`meeting_start\` without the override having been visible already.
- \`cleanup_meeting_files\` includes the new trigger so a crashed CLI invocation can't leak a stale override into the next meeting.
- \`start_meeting\` signature grew one optional param; the override is applied on top of config when present.

## Test plan

- [x] \`cargo fmt --check\` clean
- [x] \`cargo clippy --all-targets -- -D warnings\` clean (default features)
- [x] \`cargo build --features ml-diarization,onnx-load-dynamic\` builds
- [x] \`cargo test --lib meeting::\` passes (175)
- [x] \`cargo test --lib config::\` passes (85)
- [x] \`cargo test --lib --features ml-diarization,onnx-load-dynamic meeting::\` passes (180)
- [x] \`voxtype meeting start --help\` shows the new flag with possible values + env var
- [ ] Manual: start a meeting with \`--diarization simple\` on a 1:1 call, verify transcript shows \"You\" / \"Remote\"
- [ ] Manual: start a meeting with \`--diarization ml\` on a multi-speaker call, verify SPEAKER_NN labels
- [ ] Manual: omit the flag, verify config default still applies
- [ ] Manual: set \`VOXTYPE_MEETING_DIARIZATION=ml\` and confirm it's picked up

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -800,6 +800,16 @@ pub enum MeetingAction {
         /// Meeting title (optional)
         #[arg(long, short)]
         title: Option<String>,
+
+        /// Diarization backend override for this meeting only.
+        ///
+        /// `simple` attributes by audio source (You vs Remote) — best for 1:1 calls.
+        /// `ml` uses ONNX speaker embeddings for multi-speaker meetings (requires
+        /// the `ml-diarization` feature and the ECAPA-TDNN model).
+        ///
+        /// When omitted, falls back to `[meeting.diarization].backend` in config.
+        #[arg(long, value_parser = ["simple", "ml"], env = "VOXTYPE_MEETING_DIARIZATION")]
+        diarization: Option<String>,
     },
     /// Stop the current meeting
     Stop,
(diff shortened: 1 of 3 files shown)


materemias:
Addressed all four Copilot review comments in 5975edd.

### 1. CLI message misleading when diarization disabled (`src/main.rs`)
`--diarization` is a *backend* override and the daemon ignores it when `[meeting.diarization].enabled = false`. The CLI now:
- Emits a `Warning:` to stderr when `--diarization` is passed while diarization is disabled in config, explaining the override will be ignored.
- Suppresses the parenthetical suffix in that case so users don't see `(diarization: ...)` when no diarization will actually run.
- Rewords the active-case suffix from `(diarization: X)` to `(diarization backend: X)` to make the semantics explicit.

### 2. `read_trimmed_nonempty` swallowed I/O errors (`src/daemon.rs`)
Rewritten to distinguish:
- `NotFound` → `None` silently (the "no trigger file" hot path, kept quiet).
- Any other `io::Error` (permissions, partial write, transient FS) → `tracing::warn!` with the path and error, then `None`.

This way a trigger file that exists but can't be read no longer looks identical to "no file" and then gets consumed by the caller's `remove_file`.

### 3. Daemon didn't validate diarization override string (`src/daemon.rs`)
Added an `ALLOWED_DIARIZATION_OVERRIDES: &[&str] = &["simple", "ml"]` constant kept in sync with the clap `value_parser` list. The daemon now re-checks the file contents against the allowlist and drops + warns on anything else, rather than propagating an arbitrary string to `start_meeting`. The accompanying comment was updated to reflect that the daemon validates rather than relying on clap.

### 4. No clap parse tests for the new flag (`src/cli.rs`)
Added four tests inside the existing `mod tests`:
- `test_meeting_start_diarization_simple_flag` — `--diarization simple`.
- `test_meeting_start_diarization_ml_flag` — `--diarization ml --title standup`.
- `test_meeting_start_diarization_rejects_invalid` — confirms clap rejects values outside `["simple", "ml"]`.
- `test_meeting_start_diarization_env_and_default` — covers both the no-override default and the `VOXTYPE_MEETING_DIARIZATION=ml` env-var path in a single `#[test]` (env vars are process-global, so two separate tests would race when cargo runs them in parallel).

### Validation
- `cargo fmt --check` clean
- `cargo clippy --all-targets -- -D warnings` clean on default features
- `cargo test --lib cli::` (74) / `daemon::` (15) / `meeting::` (175) all pass
- `cargo test --lib --features ml-diarization,onnx-load-dynamic meeting::` (180) passes

PR #424, opened 2026-05-22
fix: support per-language XKB variants

### LINKED ISSUE #422: [Bug] [Feature] Fails to type with do-tool with non-default layout variant
## Description

Fails to type with dotool when non-default variant

## Steps to Reproduce

1. `language = ["en", "ru"]
2. restart voxtype to reload config
3. say "Пробуем говорить по-русски"

## Expected Behavior

The text spoken in Russian is typed

## Actual Behavior
The model transcribes correctly, and sends to `dotool` without setting russian layout.
The layout is not specified or specified incorrectly, so voxtype types garbage.

What works with naked `dotool`
```shell
; env XKB_DEFAULT_LAYOUT=ru XKB_DEFAULT_VARIANT=phonetic dotool
type арст
арст
```

I see there is already language_to_layout, but it is failing to be used properly. Additionally, it ignores the variant.
The variant can be set with `dotool_xkb_variant = "variant"`, but that applies to all languages, which is wrong, it should be per language.

```toml
language = ["en", "ru"]
[output.language_to_layout]
ru = "ru"
[output.language_to_variant]
ru = "phonetic"
```

## Environment

- **OS/Distro**: OpenSuse Tumbleweed latest
- **Compositor**: Plasma wayland
- **Voxtype version**: 0.7.3
- **Installation method**: .rpm
- **Whisper model**: large-v3-turbo

## Additional Context

Any other information that might help (screenshots, config file snippets, etc.)

AlexCzar:
## Summary
- add per-language XKB variant mapping alongside the existing language-to-layout mapping
- apply detected layout and variant hints to eitype and direct dotool fallback when explicit driver settings are unset
- bypass dotoolc for per-call XKB hints and document that direct dotool still requires the active desktop layout to match

## Testing
- cargo fmt
- cargo clippy --all-targets --no-deps -- -D warnings
- cargo test

Closes #422.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,11 @@ type_delay_ms = 0   # Increase if characters are dropped
 # auto_submit = true  # Send Enter after transcription (for chat apps, terminals)
 # Note: "paste" mode copies to clipboard then simulates Ctrl+V
 #       Useful for non-US keyboard layouts where ydotool typing fails
+# For multilingual layouts with variants through direct dotool fallback:
+# [output.language_to_variant]
+# ru = "phonetic"
+# dotoolc does not work with variants and cannot receive these hints. When
+# using dotool, switch your desktop layout to Russian phonetic before dictating.
 
 [output.notification]
 on_recording_start = false  # Notify when PTT activates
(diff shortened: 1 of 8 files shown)


AlexCzar:
I see two of the checks failing, but it doesn't seem like my changes' fault.

peteonrails:
@AlexCzar I agree that the failing checks are probably not sue to your changes. I'll look in to it. Thanks!

Issue #423, opened 2026-05-21
[Bug] streaming for Parakeet v3


yohans-kasaw:
## Description
installing the parakeet-tdt-0.6b-v3 model via voxtype setup model, while having [parakeet] streaming = true, fails to start.

## Steps to Reproduce

1. voxtype setup model then select parakeet-tdt-0.6b-v3 pressing 13
2. then do 'voxtype' to start it. 

## And this happens

2026-05-21T20:30:58.261478Z  INFO Loading Parakeet streaming (ParakeetUnified) model from "/home/yohansh/.local/share/voxtype/models/parakeet-tdt-0.6b-v3-int8"
Error: Transcription error: Whisper initialization failed: Parakeet streaming (ParakeetUnified) init failed: Tokenizer error: Failed to open tokenizer.model: No such file or directory (os error 2)

Streaming requires a TDT v3 model directory containing tokenizer.model.
If you're using a TDT v2 directory, switch to TDT v3 or set
[parakeet] streaming = false to use the batch pipeline.

Caused by:
    Whisper initialization failed: Parakeet streaming (ParakeetUnified) init failed: Tokenizer error: Failed to open tokenizer.model: No such file or directory (os error 2)
    
    Streaming requires a TDT v3 model directory containing tokenizer.model.
    If you're using a TDT v2 directory, switch to TDT v3 or set
    [parakeet] streaming = false to use the batch pipeline.

peteonrails:
parakeet-tdt-0.6b-v3 is not a streaming compatible model. I'm aadding an autodownload of the right parakeet model to make this simpler. In the meantime, you can grab it here: huggingface.co/bobNight/parakeet-unified-en-0.6b-onnx 

Download the 4 model files to the same directory where your other voxtype models live, then specify parakeet-unified-en-0.6b in your config and streaming should work

peteonrails:
Fixed in 0.7.4, which just merged to main. Github CI is building binaries now and the tagged release will be posted shortly

Issue #415, opened 2026-05-19
[Release process] Tarball signing step signs the wrong bytes — must sign GitHub auto-archive


peteonrails:
## Symptom

v0.7.3 shipped with a \`voxtype-0.7.3.tar.gz.asc\` release asset whose signature didn't verify against the tarball the AUR \`voxtype\` PKGBUILD downloads. \`makepkg\` users hit:

\`\`\`
==> Verifying source file signatures with gpg...
    voxtype-0.7.3.tar.gz ... FAILED (bad signature from public key 7DA2709247D75BC6)
\`\`\`

Already fixed for v0.7.3 by re-signing the auto-archive bytes and overwriting the \`.asc\` asset (AUR pushed as \`0.7.3-2\`). This issue tracks the underlying release-pipeline gap so v0.7.4 (and every release after) doesn't repeat the bug.

## Root cause

Two different tarballs exist on every GitHub release with the same canonical name \`voxtype-\$VERSION.tar.gz\`:

1. **GitHub's auto-generated source archive** at \`/archive/refs/tags/v\$VERSION.tar.gz\` — what the AUR \`voxtype\` PKGBUILD downloads (industry-standard pattern).
2. **A manually-uploaded archive** at \`/releases/download/v\$VERSION/voxtype-\$VERSION.tar.gz\` — produced by \`scripts/package.sh\` or the release-build workflow.

Different gzip metadata, possibly different archive root directory names, same source contents but different bytes. The current release workflow signs file (2) and uploads the .asc, but the PKGBUILD downloads file (1) → gpg verify fails.

## Fix forward

In \`scripts/package.sh\` (or the GH Actions release workflow), add a post-tag step:

\`\`\`bash
VERSION=...
WORK=\$(mktemp -d)
curl -sL \"https://github.com/peteonrails/voxtype/archive/refs/tags/v\$VERSION.tar.gz\" \\
  -o \"\$WORK/voxtype-\$VERSION.tar.gz\"
gpg --default-key E79F5BAF8CD51A806AA27DBB7DA2709247D75BC6 \\
    --detach-sign --armor \\
    --output \"\$WORK/voxtype-\$VERSION.tar.gz.asc\" \\
    \"\$WORK/voxtype-\$VERSION.tar.gz\"
gh release upload v\$VERSION \"\$WORK/voxtype-\$VERSION.tar.gz.asc\" --clobber
\`\`\`

Sign the auto-archive bytes; upload the resulting .asc.

[...]

Issue #406, opened 2026-05-19
[Bug] audio::levels watchdog tests are flaky under parallel cargo test (#392 follow-up)


peteonrails:
## Symptom

\`cargo test --lib\` intermittently fails on:

\`\`\`
audio::levels::tests::frame_sink_survives_respawn ... FAILED
audio::levels::tests::hub_start_accepts_a_connection ... FAILED
\`\`\`

Both pass reliably when run in isolation (\`cargo test --lib audio::levels\`). They were introduced in PR #392 (the listener watchdog/respawn fix for #391).

## Where it bites

- PR #400's CI run (the launcher-daemonize work) — \`test\` job failed on these two tests despite #400 not touching anything in \`src/audio/\`. Retrigger of just the test job hasn't been observed yet but expected to also be flaky.
- PR #405's local pre-push gate (the v0.7.3 → main merge) — same two tests fail under \`cargo test --lib\` (full parallel suite) while passing under the scoped path.

## Likely root cause

The watchdog tests create real Unix sockets. Looking at #392's approach, the tests use a process-global atomic and serialization mutex for panic injection. The socket paths themselves may not be uniquely nonced per test — two parallel tests can collide on the same path under \`/tmp\` or \`$XDG_RUNTIME_DIR\`.

## Proposed fix

In \`src/audio/levels.rs\` test module, generate a unique per-test socket path using \`tempfile::TempDir\` (already a workspace dep) or a UUID/nonce suffix. The respawn logic shouldn't share a path between independent tests.

Targeting v0.7.4 (the flake doesn't ship to users; it only burns CI cycles on PRs).

Issue #395, opened 2026-05-19
[Bug] voxtype-osd-quickshell launcher exits when invoked from a short-lived parent shell


peteonrails:
## Symptom

\`voxtype-osd-quickshell\` execs \`qs -p <dir>\` without passing \`-d\`. When invoked from a non-daemon parent (e.g. a test script, a one-shot keybinding, \`voxtype setup quickshell\` smoke test), the spawning shell exits, SIGHUP propagates through the process group, and qs terminates before its event loop settles into Wayland.

Repro:
\`\`\`
voxtype-osd-quickshell &
# qs visible briefly, then exits when the parent bash session ends
\`\`\`

Whereas \`qs -d -p /path/to/quickshell\` directly stays alive — the daemonize fork detaches from the controlling terminal so SIGHUP doesn't reach it.

## Surfaced in

v0.7.3 Quickshell visual testing tonight. Initial OSD launches survived in foreground (controlling terminal alive) but died in any background-spawn pattern (\`nohup\`, \`setsid\`, bash \`&\` + disown).

## Proposed fix

Have \`voxtype-osd-quickshell\` pass \`-d\` to \`qs\` by default. Add a \`--no-daemonize\` flag for the case where a supervisor (e.g. the daemon's OSD supervisor with \`kill_on_drop(true)\`) wants to keep the qs process attached to its child slot.

Default (daemonize): users running the launcher from a hotkey, shell, or systemd-user invocation get a persistent OSD.
Opt-out (\`--no-daemonize\`): the daemon's \`osd_supervisor\` can keep its existing \`kill_on_drop\` behavior.

Targeting v0.7.3 so the Quickshell stack works out of the box without users having to know the qs-daemonize gotcha.

peteonrails:
Retargeting to v0.7.4. v0.7.3 is now a hotfix against `main` (CPU SIGILL + a CUDA fix), not the merge of `dev`. All of the Quickshell/audio-bridge work that was scoped to v0.7.3 — including this follow-up — moves with `dev` into v0.7.4.

PR #432, opened 2026-05-26
v0.7.5: Quickshell stack, Soniox streaming, meeting-mode rework

peteonrails:
v0.7.5 is the dev-branch release that's been queued behind three hotfix slots (v0.7.2, v0.7.3, v0.7.4). It bundles the Quickshell OSD stack, the Soniox streaming backend, a substantial meeting-mode rework around diarization and audio capture, and a long tail of contributor fixes. 107 commits across 105 files (+11,755/-1,121).

It also closes the v0.7.4 streaming-Parakeet validator bug.

## Quickshell OSD stack

A native Quickshell frontend for voxtype's on-screen display, sitting alongside the existing GTK4 and native frontends. Quickshell-based desktops (Omarchy 4, end_4/dots-hyprland, custom Hyprland shells) get a first-class voxtype UI instead of having to host the GTK4 process inside their compositor.

Three widgets compose the experience:

### Waveform OSD (#381)

Canvas-rendered scrolling envelope + peak meter, ported from the Cairo `src/osd/visual.rs` implementation. Reads live audio frames from `voxtype-audio-bridge` (see below) instead of querying the daemon. Lower latency than the GTK4 frontend and no GTK runtime dependency.

### Engine picker (#385)

Quickshell menu/panel widget that lists installed ASR engines (whisper, parakeet, moonshine, sensevoice, paraformer, dolphin, omnilingual, cohere, soniox) and writes the active selection via the new `voxtype config set engine <NAME>` CLI (#382). Mirrors what the TUI Engine section does, without dropping to a terminal.

### Meeting controls (#380)

Start / stop / pause / resume buttons backed by `voxtype meeting *` commands. Active-meeting state (title, duration, chunk count) renders inline.

### Composition + installer

`shell.qml` (#387) composes the three widgets into a single shell snippet. `voxtype setup quickshell` (#390) drops the QML tree into the right shell directory and emits the import snippet for the user's shell config. The installer also symlinks `voxtype-audio-bridge` into PATH (#401) so the QML modules find it without absolute paths.

**Why use it:** the GTK4 OSD still works fine, but Quickshell users can now have a voxtype UI that lives natively in their compositor's shell layer. Omarchy 4 is moving toward Quickshell as its default, and this lands voxtype in that ecosystem.

```toml
# Enable in config.toml:
[osd]
frontend = "quickshell"
```

```bash
# Install the QML files:
voxtype setup quickshell
```

### voxtype-audio-bridge sidecar (#374)

Standalone NDJSON sidecar binary that reads `$XDG_RUNTIME_DIR/voxtype/audio.sock` and emits decoded audio frames as line-delimited JSON. Quickshell 0.2.1 doesn't expose Unix domain sockets to QML, so a tiny side-process consumed by Quickshell's `Process` element bridges the gap. Engine-agnostic, ships in every package.

### Daemonize by default in voxtype-osd-quickshell (#395)

The launcher now daemonizes by default and detaches from the supervisor so the OSD process survives daemon restarts. `--no-daemonize` is available for the supervised path. `VOXTYPE_OSD_SUPERVISED=1` flips the default at runtime.

## Soniox cloud streaming STT (#411)

Soniox is a new transcription backend that streams audio to the Soniox cloud API and returns partial+final results in real time. Joins the existing local (Whisper, Parakeet, Moonshine, SenseVoice, Paraformer, Dolphin, Omnilingual, Cohere) and HTTP (OpenAI-compatible remote) backends.

```toml
[engine]
provider = "soniox"

[soniox]
api_key = "..."
model = "stt-rt-preview"  # or whatever Soniox publishes
```

**Why use it:** voxtype's first streaming-network ASR backend. Lower latency than batch transcription for push-to-stream workflows. Drain-timeout warnings (`5s reached after end-of-audio`) point at degraded uplinks so users know when trailing finals get dropped.

Runtime requirement: the daemon's tokio runtime is multi-threaded (default for `#[tokio::main]`). When exercising `SonioxTranscriber::transcribe` from tests, use `#[tokio::test(flavor = "multi_thread")]`.

## Meeting mode

Three independent improvements landed together. Recommended to read as one feature track.

### Source diarization + audio capture fixes (#341, sjug)

Four logically separate fixes:

- `[meeting.audio].mic_device` is no longer a documented no-op — the daemon now builds a per-meeting `AudioConfig` so the override actually overrides
- `stop_meeting()` ingests trailing samples from `DualCapture::stop()` and flushes the partial chunk before the daemon stops; previously the end of every meeting silently dropped the last partial chunk
- `vad_threshold: f32` joins `[meeting.audio]` / `[meeting]` / `[meeting.chunk]` with default `0.01`, so users on quiet USB/XLR mics can lower the gate
- `SimpleDiarizer::merge_consecutive` removed: its output was 1:1 with transcript segments per the upstream `zip()` pattern, and merging silently dropped speaker labels on later segments

### VAD sub-windows + tunable ECAPA threshold (#418)

ECAPA-TDNN is trained on 2-5s utterances. Soniox-style 30s mega-segments produced averaged embeddings that failed to cluster at the old `similarity_threshold = 0.75` default. Each segment is now split into RMS-gated overlapping sub-windows (4s window, 2s hop, 0.005 RMS floor) before feeding ECAPA, and the segment's speaker is the majority vote of per-window labels. Stored centroids are updated by online mean across utterances rather than anchored on the first.

**Default change:** `similarity_threshold` moves from `0.75` to `0.25`. Only `backend = "ml"` is affected; `"simple"` and `"remote"` are unchanged.

New `[meeting.diarization]` knobs (all backward-compatible with `#[serde(default)]`):

```toml
[meeting.diarization]
similarity_threshold = 0.25   # raise to 0.30 for very-similar voices
vad_window_secs = 4.0
vad_hop_secs = 2.0
vad_rms_floor = 0.005
```

### Per-meeting --diarization override (#420)

```bash
voxtype meeting start --diarization simple   # 1:1, You vs Remote
voxtype meeting start --diarization ml       # multi-speaker
VOXTYPE_MEETING_DIARIZATION=ml voxtype meeting start
```

Diarization backend is meeting-dependent in practice: `simple` (source-keyed You/Remote) is the safer default for 1:1 calls where ECAPA over-fragments, while `ml` earns its complexity in multi-speaker meetings. The flag flips per-meeting without editing config or restarting the daemon. Reading from clap, env var, and trigger file works the way the configuration layering docs describe.

`--diarization ml` is rejected up front when the binary was built without the `ml-diarization` feature — no more silent fallback that lies about the active backend.

### Dual-source timestamp inflation fix (#330)

In dual-source meetings (mic + loopback), the meeting elapsed clock was advancing twice per audio cycle, inflating timestamps. The clock now advances once per source-synced cycle.

## Streaming Parakeet validator (closes v0.7.4 known issue)

PR #423 in v0.7.4 added `parakeet-unified-en-0.6b` (bobNight repo) as the streaming-compatible Parakeet model. The download succeeded but `validate_parakeet_model` only knew the istupakov naming (`encoder-model.onnx`, `decoder_joint-model.onnx`) and bailed on the bobNight names (`encoder.onnx`, `decoder_joint.onnx`).

v0.7.5's validator accepts both naming conventions. Users on v0.7.4 with the workaround symlinks can remove them after upgrading:

```bash
cd ~/.local/share/voxtype/models/parakeet-unified-en-0.6b
rm -f encoder-model.onnx encoder-model.onnx.data decoder_joint-model.onnx
```

## Per-language XKB variants (#424, AlexCzar)

eitype dispatch now passes XKB variant hints (e.g., `us(intl)`, `ru(typewriter)`) alongside the layout code, so non-English variants type the right characters instead of falling back to the base layout. Detected from the transcription language plus the user's configured eitype layout map.

## dotool daemon fast path (#410)

When `dotoold` is running, the dotool output method now routes through `dotoolc` (the local socket client) instead of spawning `dotool` directly. Each keystroke roundtrip drops from ~10ms to ~1ms; for long transcripts this is the difference between perceptible per-character delay and instant.

The fallback path (spawn `dotool` per call) is preserved for users without the daemon.

## wl-copy under X session (#346)

When voxtype runs under an X session (no `WAYLAND_DISPLAY`), the clipboard output method now dispatches to `xclip` or `xsel` instead of `wl-copy`, which only works on Wayland. Users running voxtype on X11 with wl-clipboard installed previously got a silent no-op.

## eitype keyboard layout hints (#180)

eitype receives the detected transcription language and uses the language→layout map in `[output.layout_hints]` to set the right `DOTOOL_XKB_LAYOUT` for the keystroke session. Mixed-language users no longer have to manually switch system layout when transcribing different languages.

## CI hardening

Several pre-release-mechanics improvements:

- **NixOS CI build job** (#369) — `nix build .#voxtype` runs on every PR and tag, catching flake regressions before they hit NixOS users
- **Parallel fmt/clippy/test** — the old monolithic `check` job is split into three parallel jobs, so a 30-second fmt failure no longer waits behind a 4-minute clippy run
- **Aggregator jobs for branch protection** — `linux-ci-success` / `macos-ci-success` / `nix-success` roll up matrix variants so adding a new build target only updates the matrix list, not the branch protection rule
- **Auto-archive signing** (#415) — already in v0.7.4, fully validated end-to-end now. The signing subkey `390AA166FE02D00BCF37150FC6277FA9D80B9EC8` lives in CI secrets and signs the GitHub-generated tarball byte-for-byte so AUR `voxtype` source builds verify cleanly
- **Hermetic test isolation** — session-detection tests (`output::session`) and audio::levels watchdog tests (#406) no longer race against each other under parallel cargo test
- **dev-branch PR template + CONTRIBUTING.md** — contributors get an explicit "PRs go to dev, mark Ready for Review only after CI is green, run `cargo fmt && cargo clippy -- -D warnings && cargo test` locally"
- **cargo clippy -D warnings enforced on PRs and main**

## Cohere model fixes (#363, jpds)

Cohere ONNX models on Hugging Face live under `onnx/` subdirs. The model registry now prefixes paths correctly for all four variants (q4f16, q4, int8, fp16). Pre-v0.7.5 setups would 404 on download.

## Audio.sock listener watchdog (#391, #392)

The audio.sock listener thread is now supervised: if the listener exits unexpectedly (orphaned by an earlier daemon crash, fd exhausted, etc.), the watchdog respawns it. Quickshell OSD users were silently losing their audio stream after long sessions.

## Bug fixes

- cuda-13 binary no longer rejects Blackwell hosts (#386) — first shipped in v0.7.4, included here for completeness
- `compiled_features()` enumerates all ONNX engines (#383) — `voxtype --version` and the TUI now report the full engine list, not just whisper + parakeet
- `MeetingButton.enabled` QML warning cleared (#399) — no more property-binding warnings on the Quickshell meeting controls
- Quickshell config layout fix (#377) so `qs` imports resolve when the QML tree is installed under `/usr/share/voxtype/quickshell/`
- Soniox documentation pinned the multi-thread runtime requirement; drain timeout logs at warn level instead of info when finals are dropped
- `__cpuid` call wrapped in explicit `unsafe` block for rustc 1.91+ compat (#419, materemias) — silent unused_unsafe warning on older toolchains via `#[allow(unused_unsafe)]`
- `CACHEDIR.TAG` written to the models directory (#367, jpds) so backup tools (restic, borg) can skip it by default
- Unused `D_MODEL` constant dropped from `cohere.rs` (jpds)
- Init-failed error prefix no longer mentions Whisper-specifically when shared across backends (#365, jpds)

## Contributors

- @sjug (meeting audio + source diarization)
- @materemias (ECAPA threshold + VAD sub-windows, --diarization flag, cpu unsafe wrap)
- @jpds (cohere onnx/ prefix, CACHEDIR.TAG, compiler warning cleanup, init-failed prefix)
- @AlexCzar (per-language XKB variants)

Plus everyone who filed issues, sent PRs, or pinged in Discussions.

## Downloads

| Variant | Binary | Use case |
|---|---|---|
| AVX2 | `voxtype-0.7.5-linux-x86_64-avx2` | Pre-Zen 4, pre-Skylake-X x86_64 CPUs |
| AVX-512 | `voxtype-0.7.5-linux-x86_64-avx512` | Zen 4+, Skylake-X+, Sapphire Rapids |
| Vulkan | `voxtype-0.7.5-linux-x86_64-vulkan` | Cross-vendor GPU acceleration for Whisper |
| ONNX AVX2 | `voxtype-0.7.5-linux-x86_64-onnx-avx2` | ONNX engines (Parakeet, Moonshine, SenseVoice, ...), CPU |
| ONNX AVX-512 | `voxtype-0.7.5-linux-x86_64-onnx-avx512` | ONNX engines, modern x86_64 CPUs |
| ONNX CUDA 12 | `voxtype-0.7.5-linux-x86_64-onnx-cuda-12` | ONNX engines, NVIDIA GPUs on CUDA 12 |
| ONNX CUDA 13 | `voxtype-0.7.5-linux-x86_64-onnx-cuda-13` | ONNX engines, NVIDIA GPUs on CUDA 13 (Blackwell) |
| ONNX MIGraphX | `voxtype-0.7.5-linux-x86_64-onnx-migraphx` | ONNX engines, AMD GPUs |
| OSD | `voxtype-0.7.5-linux-x86_64-osd`, `*-osd-gtk4` | On-screen visualizer launcher + GTK4 frontend |
| macOS | `voxtype-0.7.5-macos-universal.dmg` | arm64 Apple Silicon |

All binaries and companion `.so` libraries are detach-signed by the signing subkey. SHA256SUMS.txt and its signature are attached. To verify:

```bash
gpg --recv-keys E79F5BAF8CD51A806AA27DBB7DA2709247D75BC6
gpg --verify voxtype-0.7.5-linux-x86_64-onnx-migraphx.asc voxtype-0.7.5-linux-x86_64-onnx-migraphx
gpg --verify SHA256SUMS.txt.asc SHA256SUMS.txt
sha256sum -c SHA256SUMS.txt
```

## What's next

v0.8.0 is queued as a refactor + documentation pass shortly after v0.7.5 ships. No new features planned for the v0.8.0 slot; the goal is to tidy the heavier modules (`src/daemon.rs`, `src/config.rs`), polish the user manual, and update the troubleshooting guide before adding more surface area.

## Known issues

- **Quickshell OSD overlay captures pointer events screen-wide** (#440). When `[osd] frontend = "quickshell"` is active and voxtype is recording, the OSD's transparent fullscreen-anchored layer-shell surface eats pointer events from other windows. Text selection, cut/paste, and click-to-focus in adjacent apps stop working until the recording ends. The GTK4 frontend (the default) is unaffected. Workaround: keep `[osd] frontend = "gtk4"` until v0.7.6, which ships the proper input-region fix.

## Signing key rotation

v0.7.5 introduces a dedicated CI release signing primary, cross-signed by the existing maintainer key. The transition is transparent to AUR `voxtype-bin` users — both keys are listed in `validpgpkeys`, makepkg auto-fetches the new key from public keyservers by fingerprint on first install.

| Key | Fingerprint | Used for |
|---|---|---|
| Maintainer primary (offline) | `E79F5BAF8CD51A806AA27DBB7DA2709247D75BC6` | Git commits, manual artifact signing, cross-signing the CI key |
| CI release primary (in CI secrets) | `9CCF7915B750CAE8B095ED1AA3FC9F33FD209279` | All CI-driven signing: source auto-archive, `.deb`, `.rpm`, binary `.asc` files |

The new CI primary is standalone (not a subkey of the maintainer key) so it's auto-fetchable by full fingerprint from any public keyserver. The old signing subkey `C6277FA9D80B9EC8` used in v0.7.4 is orphaned and will be revoked once v0.7.5 sigs are verified clean — that subkey's use in v0.7.4 broke `yay -Syu` for users with stale maintainer-primary copies (#437), which v0.7.5 fixes structurally.

If you want to verify v0.7.5 sigs offline:

```bash
gpg --keyserver hkps://keys.openpgp.org --recv-keys \
    E79F5BAF8CD51A806AA27DBB7DA2709247D75BC6 \
    9CCF7915B750CAE8B095ED1AA3FC9F33FD209279
gpg --verify voxtype-0.7.5-linux-x86_64-onnx-migraphx.asc voxtype-0.7.5-linux-x86_64-onnx-migraphx
```

The Downloads section's example above also recv-keys the new primary fingerprint.


materemias:
probably tag me with the Soniox STT integration if someone needs related support

PR #433, opened 2026-05-26
feat: queue normal batch recordings

### LINKED ISSUE #28: [Feature] Cache audio input
## Problem
I like to catch mistakes early by splitting my inputs every now and then. `voxtype` doesn't accept input while processing.

## Proposed Solution
Queue consequent inputs for transcription.

AlexCzar:
## Summary

Closes #28.

Adds FIFO queueing for normal batch recordings so a user can start/stop a new dictation while a previous normal batch is still transcribing or outputting.

This touches daemon state transitions and recording/output ordering. I'm happy to adjust the shape if something here breaks intent or architectural/design preferences.

Key behavior:
- adds `[recording] queue_enabled` and `queue_size`
- keeps default queueing disabled
- keeps default `queue_size = 5`
- treats `queue_size = 0` and `queue_size = 1` as disabled
- minimum enabled queue size is `2`
- preserves FIFO processing for stopped recordings
- supports `record stop --file[=PATH]` output overrides

## Implementation Notes

- Adds a `recording_queue` module for queue lifecycle state and capacity rules.
- Moves queue invariants behind a smaller interface: live recording reservation, stopped queued jobs, processing/deferred output counts, and drain eligibility.
- Preserves cancel behavior when queued live recording state overlaps an earlier transcription.
- Prevents queue draining from overwriting live recording state while capture is still active.
- Updates CLI, config docs, default config comments, TUI guidance, user manual, and smoke-test docs.

## Compatibility

Queueing remains opt-in. Existing configs without `[recording]` keep previous single-flight behavior.

`queue_size = 0` or `queue_size = 1` disables queueing even when `queue_enabled = true`. This avoids ambiguous off-by-one behavior; queueing is active only with `queue_size >= 2`.

## Verification

- `cargo fmt --check`
- `cargo clippy --all-targets --no-deps -- -D warnings`
- `cargo test`

Manual smoke:
- ran an isolated daemon with `queue_enabled = true`, `queue_size = 2`, file output, and delayed post-processing
- recorded A, then recorded B while A was still transcribing/outputting
- verified both outputs were written in FIFO order


AlexCzar:
Rebased on latest upstream, fixed conflicts, and adjusted for the new architecture.

AlexCzar:
Hi, I already rebased once, and now there are conflicts again.
It's been a couple of months without comments, if there is no interest, please close the PR, otherwise, I can update it one last time.
Thanks.

Issue #440, opened 2026-05-27
Quickshell OSD overlay captures pointer events across the whole screen


peteonrails:
## Summary

When `[osd] frontend = "quickshell"` is active and voxtype is recording (so the Quickshell waveform OSD is visible), the OSD's fullscreen-anchored transparent layer-shell surface captures pointer events across the entire screen. This blocks text selection, click-to-focus, and drag operations in every other window for the duration of every recording — making the Quickshell OSD effectively unusable for users who type-and-talk.

The GTK4 OSD frontend (`frontend = "gtk4"`) does not have this problem; this is specific to the Quickshell stack.

## Steps to reproduce

1. Install voxtype 0.7.5 with the Quickshell stack:
   ```
   voxtype setup quickshell
   sed -i 's/^frontend = "gtk4"/frontend = "quickshell"/' ~/.config/voxtype/config.toml
   systemctl --user restart voxtype
   ```
2. Open a text editor (kate, gedit, kitty, etc.) and put some text in it.
3. Start a voxtype recording: `voxtype record start` (leave it recording).
4. While the Quickshell OSD waveform is visible at the bottom of the screen, try to:
   - Click into the text editor window
   - Select text by clicking-and-dragging
   - Right-click for a context menu
   - Ctrl+C / Ctrl+V

## Expected behavior

Pointer events pass through the OSD's transparent regions to whatever window is underneath. The OSD is a display-only indicator — it should never intercept input.

## Actual behavior

All pointer events go to the OSD layer surface and are dropped. Other windows do not receive focus changes, selection, drag, or click events. The OSD is fullscreen-anchored (transparent except for the waveform card), so the click-through region is effectively zero.

## Root cause

`quickshell/OsdSurface.qml` declares a `PanelWindow` with:

```qml
visible: daemonState !== "idle" && daemonState !== ""
anchors { top: true; bottom: true; left: true; right: true }
color: "transparent"
WlrLayershell.layer: WlrLayer.Overlay
WlrLayershell.keyboardFocus: WlrKeyboardFocus.None
```

`keyboardFocus: None` correctly suppresses keyboard input — but pointer events are a separate concern. Without an explicit `mask: Region { ... }` set to an empty input region (or to just the card's geometry), the compositor routes pointer events to this fullscreen-anchored layer instead of underlying windows.

[...]

PR #441, opened 2026-05-28
Make Soniox unconditional (ship in every v0.7.5 binary)

materemias:
## Summary

- Promotes `soniox` from a Cargo feature gate to an always-on backend, so every release binary in the v0.7.5 matrix (whisper avx2/avx512/vulkan + onnx-avx2/avx512/cuda-12/cuda-13/migraphx) can talk to Soniox out of the box.
- Aligns Soniox with the existing OpenAI-compatible `RemoteTranscriber`, which has always shipped unconditionally — same shape (pure WebSocket + HTTP client, no model, no GPU), so the asymmetric gating was the odd one out.
- Drops the corresponding `#[cfg(feature = "soniox")]` gates and the "Requires: cargo build --features soniox" doc lines.

## Why now

This is queued behind #432 (v0.7.5). The Quickshell engine picker added in that PR lists Soniox alongside the local engines; if some release flavors lack the backend, the picker either lies per binary or has to grow capability detection. Shipping Soniox in every binary keeps the engine surface uniform across the release matrix.

## Lightweight by design — should just work

Soniox is the lightest backend voxtype has: no local model, no GPU, no CPU-instruction concerns, no per-distro shared library. It's a thin WebSocket + HTTP client. There is no hardware or platform reason for it to need a special build — the only thing standing between a user and Soniox is the API key. Putting that behind `--features soniox` makes the lightest backend the *hardest* one to enable from a fresh install (you have to find the docs, learn cargo features, rebuild). Flipping it to always-on means: install any voxtype binary, drop an API key in the config, done. That matches the "dead simple user experience" principle in `CLAUDE.md`.

## Cost

`tokio-tungstenite` (rustls) + `futures-util` + `reqwest` (rustls) become required deps. After LTO + strip this adds roughly 1-2 MB on top of the `ureq` client every binary already carries for model downloads. Whisper-only binaries pay it too, but they already ship `ureq` for setup downloads — the "fully offline" ship has already sailed.

## Files

- `Cargo.toml` — drop `soniox` feature, demote `optional = true` on the three deps.
- `src/transcribe/mod.rs` — remove module-decl gate and the two factory-arm gates (the `#[cfg(not(feature = "soniox"))]` "engine requested but not compiled in" arm goes away with them).
- `src/config.rs` — drop two stale "Requires: cargo build --features soniox" doc lines on `SonioxConfig` and `TranscriptionEngine::Soniox`.
- `docs/CONFIGURATION.md`, `docs/USER_MANUAL.md`, `docs/SONIOX.md`, `docs/MODEL_SELECTION_GUIDE.md` — update build instructions and the feature-matrix entry.

## Target release

**v0.7.5** — should land before #432 ships so the engine surface is uniform from the first v0.7.5 binary onward.

## Test plan

- [x] `cargo build --release` (no `--features soniox`) succeeds on dev branch + this change
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --release -- -D warnings` clean
- [x] `cargo test --release --lib soniox` — 41 passed
- [x] `cargo test --release --lib` — only pre-existing `setup::parakeet::tests::detect_available_backends_returns_vec` failure (environment-dependent, reproduces on dev without this change)
- [ ] Confirm Soniox actually transcribes end-to-end with a release binary built without `--features soniox`
- [ ] Verify Quickshell engine picker shows soniox on a whisper-only binary (no ONNX engines)

## Out of scope

- Updating the `Dockerfile.onnx*` `--features` lists. They no longer need `soniox` appended (the feature is gone), and they're currently feature-clean of it already, so no Dockerfile edits are required for this PR. The MIGraphX / CUDA Dockerfiles are unaffected.
- The Nix flake — same reasoning. If the flake passes `--no-default-features`, soniox keeps working because its deps are no longer behind a feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,11 +55,11 @@ hound = "3"  # WAV file reading/writing
 # HTTP client for remote transcription
 ureq = { version = "2", features = ["json"] }
 
-# WebSocket client for Soniox streaming backend (optional)
-tokio-tungstenite = { version = "0.24", optional = true, features = ["rustls-tls-webpki-roots"] }
-futures-util = { version = "0.3", optional = true }
+# WebSocket client for Soniox streaming backend
+tokio-tungstenite = { version = "0.24", features = ["rustls-tls-webpki-roots"] }
+futures-util = "0.3"
 # Async HTTP client for Soniox async transcription API (multipart upload + poll)
-reqwest = { version = "0.12", optional = true, default-features = false, features = ["json", "multipart", "rustls-tls"] }
+reqwest = { version = "0.12", default-features = false, features = ["json", "multipart", "rustls-tls"] }
 
 # JSON parsing (for CLI backend)
 serde_json = "1"
@@ -238,8 +238,10 @@ omnilingual-tensorrt = ["omnilingual", "onnx-tensorrt-enabled"]
 cohere = ["onnx-common", "dep:half", "dep:tokenizers"]
 cohere-cuda = ["cohere", "onnx-cuda-enabled"]
 cohere-tensorrt = ["cohere", "onnx-tensorrt-enabled"]
-# Soniox cloud streaming WebSocket STT backend (no local model, just a network client)
-soniox = ["dep:tokio-tungstenite", "dep:futures-util", "dep:reqwest"]
+# Soniox is unconditional. Its deps (tokio-tungstenite, futures-util, reqwest)
+# add ~1-2 MB after LTO+strip and ride alongside ureq, which every binary
+# already ships for model downloads. Treating it like remote.rs (also always
+# compiled) keeps the engine surface uniform across release flavors.
 # No cohere-migraphx feature: MIGraphX 7.2 still fails on the
 # onnx-community q4 export. Original blocker was MatMulNBits bits=8
 # (cstr int8); q4 has bits=4 which is supported, but its zero_points
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -70,7 +70,7 @@ config changes; restart it with `systemctl --user restart voxtype` for the
 new engine to take effect.
 
 **Notes:**
-- All engines except Whisper require an ONNX-enabled binary (`voxtype-*-onnx-*`)
+- Whisper, Remote Whisper, and Soniox run in every binary. The other engines (Parakeet, Moonshine, SenseVoice, Paraformer, Dolphin, Omnilingual, Cohere) require an ONNX-enabled binary (`voxtype-*-onnx-*`)
 - Each ONNX engine reads its own `[<engine>]` section (e.g. `[parakeet]`, `[cohere]`)
 - See [PARAKEET.md](PARAKEET.md) for detailed Parakeet setup instructions
 - See [MOONSHINE.md](MOONSHINE.md) for detailed Moonshine setup instructions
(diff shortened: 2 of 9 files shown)


materemias:
**CI status:** code checks (`fmt`, `clippy`, `test`, `ci-success`) all green.

The `build (cpu)`, `build (onnx)`, and `nix-success` failures are the ongoing Nix fetchCrate 403 wave that's red on every dev-targeted PR right now (including #432). Root cause: `curl: (22) The requested URL returned error: 403` against the crates.io tarball mirror, repeating across `accesskit`, `base64ct`, `bit-set`, `calloop`, etc. — same crates fail in identical order regardless of branch. Not caused by this Cargo.toml change. Will re-kick once the upstream is sorted.

materemias:
Addressed Copilot's review (pushed as `6a1c55b`):

- `docs/MODEL_SELECTION_GUIDE.md:5` — intro now reads "eight transcription engines. Three ship in every binary — Whisper (local), Remote Whisper (HTTP API), and Soniox (cloud streaming). The other five require the ONNX binary variant." Matches the "Built-in" cell in the table below.

While I was in the docs, I swept two adjacent staleness cases the same Copilot finding implied:

- `docs/CONFIGURATION.md:73` — the `voxtype config set engine` Notes block previously said "All engines except Whisper require an ONNX-enabled binary." Updated to call out Whisper, Remote Whisper, and Soniox as bundled, and ONNX-only engines explicitly listed.
- `docs/TROUBLESHOOTING.md:1110` — the 408 fix paragraph said "any release after v0.7.2 + soniox" (the "+ soniox" qualifier meant `--features soniox`). Now reads "any release v0.7.3 or later includes it" since every release v0.7.5+ ships Soniox unconditionally.

materemias:
Correcting myself on the previous comment: pushed `b9d5626` fixing the 408 fix cutoff from "v0.7.3 or later" to "v0.7.5 or later" in `docs/TROUBLESHOOTING.md`. The fix code has been there since v0.7.3 but only reachable from soniox-feature-flag builds; v0.7.5 is the first release where every binary universally has both the fix and a working Soniox path.

peteonrails:
Thanks for putting this together @materemias: the reasoning is solid 

**Queued for v0.7.6.**

Why not v0.7.5: rc/0.7.5 is in the final stretch waiting on the FlakeHub cache issue. I've shipped four last-minute commits into rc/0.7.5 today already (CI release primary key rotation, .deb/.rpm in CI, voxtype-osd-quickshell + audio-bridge packaging, and the #443 wrapper fix). 

The v0.7.6 release is queued as a focused bug-fix cycle (closing #440, the streaming-on-incompatible-model startup error from #442, and the ORT EP runtime-status honesty from #443). This PR fits cleanly as a UX-polish addition alongside those

Once v0.7.5 ships and the standard `main → dev` post-release merge lands, dev will be ready to take this PR. The merge should be fast-forward at that point without copnflicts

One remaining test-plan item before merge into dev:  the unchecked "Confirm Soniox actually transcribes end-to-end with a release binary built without `--features soniox`"

I'll cover that locally when we kick off v0.7.6 by building a release binary against this branch and running a streaming session against the Soniox cloud. Unit tests passing is good signal, but I'll run a real-world smoke test just to be sure

materemias:
Sure. By the way, I found Soniox to be the most pleasant to use among all the models I tried for the languages I dictate, largely because of its high-quality transcription and fast processing speed. I hope you will like it too.

PR #476, opened 2026-05-30
Split src/main.rs into src/main.rs + src/app/ folder (SSoT cleanup follow-on)

peteonrails:
## Summary

Mirrors the [config split (PR #469)](https://github.com/peteonrails/voxtype/pull/469) and the parallel-branch [cli split (PR #473)](https://github.com/peteonrails/voxtype/pull/473). `src/main.rs` went from 2571 lines to a 75-line shim; the rest moved into 14 files under `src/app/` organised by subcommand. Then five SSoT dedup wins follow on top, several of which target the canonical examples in [`docs/REFACTORING.md`](https://github.com/peteonrails/voxtype/blob/dev/docs/REFACTORING.md).

Plan note for full context: `~/Documents/markdown-notes/Voxtype/2026-05-30-main-cli-refactor-plan.md`.

## Commits

The PR is four stacked commits, each independently green; review them in order if it helps.

### 1. `Split src/main.rs into src/main.rs + src/app/ folder`

Mechanical move. `fn main()` stays in `src/main.rs` (tokio attribute, sigill install, sigpipe reset, log init, `Cli::parse()`, config load) and calls `app::run`. The handler bodies, the 290-line CLI override block, and the dispatch match all move into `src/app/{record,status,meeting,info,config_show,config_set_engine,transcribe_file,updates,macos,sigpipe,overrides,dispatch}.rs`. Three orchestrators are new (`app::run`, `apply_cli_overrides`, `dispatch::dispatch`); the rest is text moves. Three regression tests move next to the code they pin.

Justification: Rules 3 and 4 from REFACTORING.md (`main.rs` no longer predicts its contents — an audio resampler, a Waybar JSON formatter, a meeting CLI; every new flag adds another override stanza).

### 2. `Dedupe engine parsing, paired flags, JSON building, daemon liveness`

Five SSoT wins now that the new file shape makes the patterns visible:

- **Engine-string parsing.** The 9-arm `match engine.to_lowercase()` block was duplicated 3 places (env var loader, top-level `--engine`, `transcribe --engine`). Replaced with `engine.parse::<TranscriptionEngine>()` via new `strum::EnumString` + `ascii_case_insensitive`.
- **Paired `--foo`/`--no-foo` overrides.** 7 sites of `if cli.x { ... = true } if cli.no_x { ... = false }` collapsed to single-line `apply_bool_override` calls. Mirrors `override_from_flags` from the cli split (commit dece381).
- **`format_state_json`.** Now uses `serde_json::to_string` for value escaping (fixes a latent injection class — a device or model name containing `\"` or `\\` could produce malformed JSON). Outer template stays hand-rolled to preserve Waybar's whitespace shape. Byte-exact pin test added before the swap.
- **Daemon liveness.** The canonical SSoT example from REFACTORING.md. New `src/daemon_status.rs` lib module unifies 5 sites; the TUI's copy used the wrong filename (`runtime_dir/pid` instead of `voxtype.lock`) and was silently reporting every modern build's daemon as down.
- **Regression test** for `pid_file_path_matches_send_record_command` moved into the new lib module.

### 3. `Apply reviewer feedback: bound PIDs, complete liveness SSoT, fix nits`

Three reviewers (code, architecture, security) ran against commits 1+2. This commit applies their feedback:

- **Security MEDIUM:** reject `pid <= 1` in `daemon_status::read_pid` and `check_daemon_running`. `libc::kill` overloads non-positive PIDs (`0` = process group, `-1` = broadcast, `< -1` = signal a process group). A corrupted lockfile reading `-1\\n` would have made `voxtype record start` mass-signal the user's session.
- **Code review:** add `conflicts_with = \"audio_feedback\"` to `--no-audio-feedback` (the only paired flag lacking it; the new `apply_bool_override` helper's `enable`-wins precedence differed from the old two-`if`-block disable-wins precedence).
- **Code review:** add a byte-exact pin assertion for `\"` and `\\` escaping — the motivating case for the serde_json switch. Round-trips through `serde_json::Value`.
- **Code review:** fix stale docstring in `src/app/mod.rs` that still referenced the deleted `daemon_pid.rs`.
- **Bonus SSoT completion:** fold `daemon.rs::is_pid_running` into `daemon_status::is_running`. The Linux implementation sent `SIGCONT` to the PID before falling back to signal 0, delivering a real signal to whatever process held a recycled PID. The unified helper uses signal 0 only. Drops the only `nix::Pid`/`nix::Signal` imports in `daemon.rs`.

### 4. `Move Waybar status JSON contract into the library`

Architecture review flagged that `format_state_json` and `ExtendedStatusInfo` lived in binary-only code (`src/app/status.rs`) even though the JSON shape is the external API voxtype exposes to every status consumer (Waybar via `voxtype status --follow`, Quickshell, the planned audio bridge, a hypothetical future StatusNotifierItem tray). Moved to `voxtype::status_json` so any future in-process consumer can emit the same contract. Only `run_status` (the inotify follow loop) stays binary-side.

## What this PR is *not* doing

Per REFACTORING.md and the plan note's \"not in scope\" list, the 290-line override block in `app/overrides.rs` keeps its `if let Some(...)` shape. The axis of variation isn't stable enough to extract a derive macro or builder yet. Land the move first, revisit dedup after the new file's pattern is visible across the codebase.

## Verification gates

All gates clean on the rebased branch:
- `cargo test --lib`: 775 passed
- `cargo test --bin voxtype`: 1 passed (the 4th from earlier commits moved to lib with format_state_json)
- `cargo clippy --all-targets --no-deps -- -D warnings`: clean
- `cargo fmt --check`: clean
- `voxtype --help` for all 9 subcommands: byte-identical to the baseline captured before commit 1 (also re-verified post-rebase against the new `origin/dev`)

The only intentional user-facing behaviour change: `voxtype --audio-feedback --no-audio-feedback` now errors at parse time (\"argument cannot be used with...\") instead of silently picking one direction. Matches the six other paired flags.

## Test plan

- [ ] Existing CI (`fmt`, `clippy`, `test`) passes
- [ ] Smoke-test `voxtype daemon` start/stop with the new code path
- [ ] Smoke-test `voxtype status` (one-shot and `--follow`) — JSON shape must match what Waybar configs expect
- [ ] Smoke-test `voxtype record start/stop/toggle/cancel` — verify the daemon receives signals correctly via the new liveness path
- [ ] Smoke-test `voxtype meeting start/stop` — same liveness path with `check_daemon_running()?` discarding the returned PID
- [ ] Smoke-test `voxtype configure` (TUI) — the daemon-running banner now uses the unified helper
- [ ] Manual: stale-lockfile cleanup still works (daemon detects + removes lockfile from a crashed predecessor)

## Follow-ups noted but not in this PR

- Build.rs restructure to make `ENGINE_NAMES_CSV` derivable from `TranscriptionEngine::VARIANTS` rather than hand-maintained-with-test-pin. The current arrangement (PR #475) is constraint-imposed by build.rs's `#[path]` inclusion. A true SSoT would require restructuring build.rs's man-page generation.
- Architecture review flagged that `apply_cli_overrides` reaches into `setup::model::is_valid_model` and `notification::send_sync` — couples CLI parsing with desktop session. Tractable, single call site, defer until there's a second.
- Architecture review noted `app/` is a generic name; `commands/` or `cli_main/` would predict contents better. Rename costs a `super::` path update in every submodule; not worth it pre-merge.


peteonrails:
Two follow-ups noted but deferred to separate PRs per scope discipline:

- #477 — Simplify `apply_cli_overrides`: split into per-section helpers + add `strum::EnumString` to `WhisperMode`/`FileMode`/`VadBackend`. ~80 line reduction, behaviour-preserving.
- #478 — Restructure `build.rs` to derive `ENGINE_NAMES_CSV` from `TranscriptionEngine::VARIANTS` (compiler-enforced SSoT rather than test-pinned). Achievable via a `variants.rs` leaf module split; ~5s build cost.

PR #494, opened 2026-06-09
Fix dropped characters in streaming output during self-corrections

materemias:
## Problem

During Soniox streaming dictation, typing intermittently drops a character — most often a **space**, and most noticeably **during self-corrections**.

The streaming reconciler revises non-final tokens by backspacing the previously-typed tail and retyping it. Those backspaces were emitted with **no inter-key pacing**. On KWin (which coalesces uinput events arriving faster than one display frame, ~16.67ms at 60Hz):

- rapid backspaces got coalesced and dropped → a stray leftover character, and
- the un-spaced last backspace let the following retype's leading character (usually a space) coalesce too.

The **type** path was already safe — `DotoolOutput::build_commands` sets `typedelay`/`typehold` from `type_delay_ms`. Only the **backspace** path was unprotected. Crucially, dotool's `key` events obey `keydelay`/`keyhold`, **not** `typedelay`/`typehold` (which only apply to `type`), and those default to ~2ms.

## Fix

Thread the configured `type_delay_ms` into `StreamingSession` and pace all three backspace backends:

| Backend | Pacing |
|---|---|
| dotool | `keydelay N` + `keyhold N` before the `key backspace` lines |
| wtype | `-s N` before each `-k BackSpace` (wtype's `-d` only paces text typing, **not** discrete `-k` presses) |
| ydotool | `key -d N …` |

When `type_delay_ms == 0` (user opted out), no pacing is added on any backend — behavior is unchanged.

## Also in this PR

- Extracted `build_dotool_backspace_commands` / `build_wtype_backspace_args` as pure, unit-tested helpers.
- Removed the unused `Default for StreamingSession` impl — it constructed with delay `0`, silently disabling pacing (a latent footgun).
- Gitignored the local `.codegraph/` runtime state.

## Testing

- `cargo build --features soniox` ✅
- `cargo test --features soniox --lib output::streaming` — 11/11 ✅ (5 new helper tests covering keydelay/keyhold vs typedelay, zero-delay opt-out, zero-count, and wtype `-s`-per-key)
- `cargo fmt --all --check` ✅
- `cargo clippy --all-targets -- -D warnings` on both default and `--features soniox` ✅

## Known follow-up (out of scope)

`emit_backspaces` uses a hardcoded `wtype → dotool → ydotool` order independent of the user's `[output] driver_order` (the type path respects it via `output_with_fallback`). This predates this change — the fix just paced each existing backend. Unifying the backspace path with `driver_order`/the `TextOutput` trait is a larger refactor worth a separate issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ website/docs
 # Local agent worktrees and ephemeral plans
 .claude/worktrees/
 PLAN-*.md
+
+# CodeGraph index + machine-local runtime state (daemon.pid, sockets)
+.codegraph/
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -892,7 +892,7 @@ impl Daemon {
 
         *audio_capture = Some(capture);
         *streaming_handle = Some(handle);
-        *streaming_session = Some(StreamingSession::new());
+        *streaming_session = Some(StreamingSession::new(self.config.output.type_delay_ms));
         *streaming_chain = Some(output::create_output_chain(&self.config.output));
         *state = State::Streaming {
             started_at: std::time::Instant::now(),
--- a/src/output/streaming.rs
+++ b/src/output/streaming.rs
@@ -67,15 +67,26 @@ pub struct StreamingSession {
     typed_chars: usize,
     /// Most recent partial text (for status only; never typed).
     partial: String,
+    /// Inter-key delay (ms) applied to backspace bursts, mirroring the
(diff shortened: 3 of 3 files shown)


peteonrails:
@materemias I am doing a deep dive test on 1.0.0 at the moment, will review and  look to merge this for 1.0.1!

materemias:
Following up now that v1.0.0 is released. CI is green and the pacing fix is still relevant. Is this still a candidate for v1.0.1?

PR #503, opened 2026-06-16
fix: improve input startup and streaming output reliability

materemias:
## Summary
- warn once when a Voxtype dotool XKB override forces direct dotool despite a running dotoold, with separate guidance for layout-only and variant overrides
- pace streaming backspace bursts using the configured output delay to avoid dropped/coalesced replacement events
- add audio device warm-up gating (wait_for_device) so recording-start cues wait for real input samples and reduce missed first words

## Verification
- cargo fmt --check
- cargo test dotool

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ website/docs
 # Local agent worktrees and ephemeral plans
 .claude/worktrees/
 PLAN-*.md
+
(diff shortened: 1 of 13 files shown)


materemias:
Superseded by the focused PRs: #494 for streaming backspace pacing, #497 for audio-device warm-up, and #672 for the dotool XKB slow-path warning. Closing the combined branch to avoid duplicate review.

PR #504, opened 2026-06-16
feat: update Soniox models to v5

materemias:
## Summary

Updates the default Soniox models from v4 to v5:

- Realtime WebSocket default: `stt-rt-v4` → `stt-rt-v5`
- Async REST default: `stt-async-v4` → `stt-async-v5`

The async-default swap in `SonioxTranscriber::new` now recognizes both the old (`stt-rt-v4`) and new (`stt-rt-v5`) realtime defaults, so a user who had the v4 default in their config still gets transparently routed to `stt-async-v5` when `async_api = true`. Explicit non-default model overrides are left untouched.

## Changes

- `src/config/engines/soniox.rs` — default model `stt-rt-v5`
- `src/transcribe/soniox.rs` — async-default swap matches v4 or v5 realtime, swaps to `stt-async-v5`; tests updated
- `src/config/root.rs` — meeting-mode routing log and doc comment reference v5
- `docs/CONFIGURATION.md`, `docs/SONIOX.md`, `docs/MODEL_SELECTION_GUIDE.md` — model names updated

## Backwards compatibility

Configs pinned to the old default (`stt-rt-v4`) continue working: the realtime path uses whatever is configured, and the meeting/async path swaps the v4 *default* to the v5 async model the same way it always did for v4. No config field changes.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1409,10 +1409,10 @@ export SONIOX_API_KEY="your-key-here"
 ### model
 
 **Type:** String
-**Default:** `"stt-rt-v4"`
+**Default:** `"stt-rt-v5"`
 **Required:** No
 
-Soniox model identifier. The current realtime model is `stt-rt-v4`.
+Soniox model identifier. The current realtime model is `stt-rt-v5`; when `async_api = true`, voxtype uses `stt-async-v5` unless you explicitly override `model`.
 
 ### language_hints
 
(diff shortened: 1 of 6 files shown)


materemias:
Thanks @copilot-pull-request-reviewer — addressed in 58cfbab. All five points came down to the same subtlety: the swap routes *realtime-default* values (`stt-rt-v4`/`stt-rt-v5`) to `stt-async-v5`, while any other explicit `model` is kept as-is.

- **soniox.rs** — added `async_api_swaps_legacy_v4_realtime_default`, a regression test for the `stt-rt-v4` → `stt-async-v5` backwards-compat path.
- **CONFIGURATION.md** — `### model` and the `async_api` notes now spell out that a realtime-default `model` is swapped while other explicit values are used as-is, instead of implying any override blocks the swap.
- **root.rs** — dropped the hard-coded `stt-async-v5` from the meeting-mode log line (the configured model can differ under an explicit override); doc comment now says "default model `stt-async-v5`".
- **SONIOX.md** — updated the quoted log string to match the new, model-agnostic message.

fmt, clippy (`-D warnings`), and the soniox/meeting-mode lib tests all pass locally.

PR #505, opened 2026-06-16
fix: resolve Soniox model per-mode instead of swapping (Option<String>)

materemias:
Stacked on #504 (base: `feature/soniox-v5`). Retarget to `dev` once #504 merges.

## Problem

Soniox has a single `model` config field shared by the realtime WebSocket and async REST paths. The swap heuristic (`stt-rt-*` → `stt-async-v5` under `async_api`) couldn't distinguish "user explicitly set `stt-rt-v5`" from "user left the default" — both are the same string — so an *explicit* realtime pin still got silently rewritten under `async_api`/meeting mode. The docs had to warn about this surprise.

## Change

Make `model` an `Option<String>` (None = unset) and resolve at the use site:

- **Unset** → per-mode default: `stt-rt-v5` (realtime) or `stt-async-v5` (async, including meeting mode, which always uses async).
- **Explicit** → used verbatim in both modes, no auto-swap.

A single `resolved_model()` method centralizes this; the swap block in `SonioxTranscriber::new` is gone, and the WS init frame, async request body, logging, and `Config::model_name()` all route through it. No second config field added.

## Backwards compatibility

- Default configs are unaffected: `None` still resolves to `stt-async-v5` in meeting mode.
- `None` is omitted from serialized TOML (verified by test), so nothing writes an explicit `model` into a user's config. `save_config` is dead code, `config_set` does comment-preserving surgical edits, and the shipped template has no `[soniox] model` line — so a `model` value is present only if hand-typed.
- **One behavior change:** a user who *explicitly* pinned `model = "stt-rt-v5"` (or `stt-rt-v4`) *and* uses async/meeting mode previously got auto-swapped to `stt-async-v5`; now the explicit pin is honored as-is. Rare, and arguably the more correct reading of an explicit override.

## Tests

- New resolver tests in `config/engines/soniox.rs` (run without the `soniox` feature): unset→realtime, unset+async→async, explicit-verbatim-in-both-modes, and `None`-omitted-from-TOML.
- Updated transcriber tests for the new semantics.
- fmt, clippy (`-D warnings`), build (default + `soniox`), and lib tests all pass locally.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1508,7 +1508,7 @@ terms_file = "/home/me/dotfiles/voxtype-terms.json"
 **Default:** `false`
 **Required:** No
 
-Use the Soniox **async transcription API** (file upload + poll) instead of the realtime WebSocket. Different model (`stt-async-v5`), different accuracy profile, batch only — no live partials, no flicker, push-to-talk compatible.
+Use the Soniox **async transcription API** (file upload + poll) instead of the realtime WebSocket. Different model (`stt-async-v5` by default, unless `model` is set explicitly), different accuracy profile, batch only — no live partials, no flicker, push-to-talk compatible.
 
 When `true`:
 - Audio buffered while recording. On release, voxtype uploads the WAV to `https://api.soniox.com/v1/files`, creates a transcription job, polls until complete, fetches the transcript, then types it at the cursor in one shot.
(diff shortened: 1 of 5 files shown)


materemias:
Thanks @copilot-pull-request-reviewer — both addressed in d33f659.

- **soniox.rs** — the TOML-omission test now asserts on `model =` specifically, so a future field like `model_type` won't trip it.
- **CONFIGURATION.md** — the `async_api` "Different model" sentence now says `stt-async-v5` is the default *unless `model` is set explicitly*, matching the new verbatim-override behavior.

CI is green and fmt/clippy (`-D warnings`)/tests pass locally.

PR #510, opened 2026-06-19
Fix remote Whisper model and Unicode output fallback

### LINKED ISSUE #496: [Bug] Remote mode: remote_model config ignored, falls back to model (local model name) causing 404 on OpenAI-compatible APIs
Environment:
- voxtype version: 0.7.5 (paru/Arch)
- OS: Arch Linux, KDE Plasma Wayland
- Engine: whisper, mode: remote, endpoint: https://api.groq.com/openai
- Config: remote_model = "whisper-large-v3-turbo", model = "large-v3-turbo"

Problem:
In mode = "remote", voxtype ignores the remote_model setting and sends the local model value to the remote API endpoint. This causes a 404 error because local model names (e.g., large-v3-turbo) don't exist on remote OpenAI-compatible APIs (Groq, OpenAI, etc.) which expect names like whisper-large-v3-turbo.

Observed behavior:
1. Config has:
[whisper]
mode = "remote"
remote_endpoint = "https://api.groq.com/openai"
remote_model = "whisper-large-v3-turbo"
model = "large-v3-turbo"  # local model name
2. Log shows: Configured remote transcriber: endpoint=https://api.groq.com/openai, model=large-v3-turbo, timeout=30s
3. Remote API returns 404: The model 'large-v3-turbo' does not exist or you do not have access to it.
4. Only workaround: set model = "whisper-large-v3-turbo" (duplicate value) so remote gets correct name
Expected behavior:
When mode = "remote", voxtype should use remote_model for the API request. The model field should only apply to local backend.

Config demonstrating the issue:
[whisper]
mode = "remote"
remote_endpoint = "https://api.groq.com/openai"
remote_model = "whisper-large-v3-turbo"  # ignored
model = "large-v3-turbo"  # incorrectly sent to remote API

Impact:
- Users migrating from local to remote get confusing 404 errors
- Documentation doesn't clearly state that model must also be set to the remote model name as workaround
- Inconsistent with config structure where remote_* prefix suggests remote-specific settings

Suggested fix:
In remote mode, prioritize remote_model over model for the API request. Fall back to model only if remote_model is not set (for backward compatibility).

poommmes-del:
## Summary

Fixes the remote Whisper/Groq path from issue #496 and the follow-up output failure found while validating it with Arabic text.

## What Changed

- Preserve `[whisper].remote_model` when remote mode is used without a per-recording model override.
- Keep explicit `--model` / model modifier behavior working by only overriding `remote_model` when the caller actually supplies a runtime model.
- Avoid direct `dotool`/`ydotool` text typing for non-ASCII output, because those drivers are layout-dependent and can report success while inserting nothing or wrong characters.
- Insert paste-mode fallback before clipboard-only fallback in `mode = "type"`.
- Allow paste mode to send the paste shortcut through `dotool key ctrl+v`, which covers compositors that reject `wtype` and systems without a running `ydotoold`.

## Root Cause

`ModelManager::get_prepared_transcriber(None)` converted the absence of a runtime model override into `config.model`. Remote mode then wrote that local model name into `remote_model`, so OpenAI-compatible APIs like Groq received the wrong model.

After fixing that, Groq returned correct Arabic text, but the output chain stopped at direct `dotool` typing. `dotool` and `ydotool` synthesize layout-dependent key events, so they are not reliable for Unicode text when the active layout/compositor cannot produce those characters. The new path uses clipboard paste for non-ASCII text while still sending the paste keystroke automatically.

Closes #496.

## Validation

- `cargo fmt --check`
- `cargo test` (834 passed)
- Manual Groq validation with:
  - `remote_endpoint = "https://api.groq.com/openai"`
  - `remote_model = "whisper-large-v3-turbo"`
  - Arabic transcription on a compositor without virtual-keyboard protocol support
  - observed output path: skip direct `dotool`/`ydotool`, `wtype` fails, paste succeeds via `dotool key ctrl+v`

### DIFF
--- a/src/model_manager.rs
+++ b/src/model_manager.rs
@@ -83,9 +83,10 @@ impl ModelManager {
             return self.get_transcriber(None);
         }
 
-        // For remote backend, create transcriber with model override
+        // For remote backend, preserve configured remote_model unless the
+        // caller supplied an explicit runtime model override.
         if self.config.effective_mode() == WhisperMode::Remote {
-            return self.create_remote_transcriber(&model_name);
+            return self.create_remote_transcriber(model);
         }
 
         // For CLI backend, create transcriber each time (no caching needed)
@@ -102,14 +103,15 @@ impl ModelManager {
         self.get_or_load_cached(&model_name)
     }
 
-    /// Create a remote transcriber with model override
+    /// Create a remote transcriber with optional runtime model override
     fn create_remote_transcriber(
         &self,
-        model: &str,
+        model_override: Option<&str>,
     ) -> Result<Arc<dyn Transcriber>, TranscribeError> {
         let mut config = self.config.clone();
-        // Override remote_model with requested model
-        config.remote_model = Some(model.to_string());
+        if let Some(model) = model_override {
+            config.remote_model = Some(model.to_string());
+        }
         let transcriber = transcribe::remote::RemoteTranscriber::new(&config)?;
         Ok(Arc::new(transcriber))
     }
@@ -310,8 +312,9 @@ impl ModelManager {
             return Ok(prepared.transcriber);
         }
 
-        // No prepared transcriber, get normally
-        self.get_transcriber(Some(&model_name))
+        // No prepared transcriber, get normally. Preserve None so remote mode
+        // can use whisper.remote_model instead of the local default model.
+        self.get_transcriber(model)
     }
 
     /// Get the list of currently loaded models (for debugging/status)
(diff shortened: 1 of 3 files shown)


peteonrails:
Half of this shipped and half did not

"fix: preserve remote Whisper model" is in the 1.0.1 release branch as `7b9b28c`, with your authorship preserved, and closes #496. It ships in v1.0.1. Remote mode falling back to the local model name and 404ing against OpenAI-compatible APIs was a real blocker for anyone pointing voxtype at a remote endpoint.

**Not yet:** `309e154` "fix: paste Unicode text when typing is unsafe" has not been taken. It is 260 lines across `src/output/mod.rs` and `src/output/paste.rs`, in the output driver chain, and it arrived while 1.0.1 was being tagged. Putting a change of that size into the typing path without proper review, hours before a release, is not a good trade for either of us, so I'll incorporate it later. 

The Unicode work will be reviewed on its own merits for 1.1.x, where it can get the attention it deserves.

The 1.0.1 work was assembled on a release branch rather than merged PR by PR, which is why the first commit reached the branch by cherry-pick instead of this PR being merged. Thanks for both fixes.

PR #513, opened 2026-06-25
fix(soniox): free uploaded files on async cleanup to stop org file-cap leak

materemias:
## Problem

Meeting transcription via the Soniox async backend silently freezes mid-recording (observed: a 55-minute meeting whose transcript stopped dead at 9:00, audio not retained → unrecoverable). Root cause: the async path uploads each chunk as a Soniox **file**, but cleanup only deleted the **transcription**, on the belief that `DELETE /v1/transcriptions/{id}` cascades to the file.

It does not. Confirmed empirically — the account sat at **999 files / 1 transcription** — and by Soniox's own async example, which deletes both resources explicitly. Every successful chunk therefore leaked its file until the org-wide **1,000-file cap** was hit, after which every `POST /v1/files` returns `429 limit_exceeded` and transcription stops. The transcriptions cap (2,000) has more headroom, so cleanup was draining the abundant resource while leaking the scarce, binding one.

## Fix

- **Delete the file, not just the transcription.** `async_cleanup` now deletes the transcription first, then the uploaded file (the order Soniox's example uses); `file_id` is threaded into cleanup.
- **Cleanup runs unconditionally.** Poll + fetch are wrapped in an async block so every exit — including `?` early-returns on transient transport/parse errors during poll/fetch — falls through to a single cleanup call. Previously those `?` paths returned early and leaked both resources.
- **Retry transient deletes.** Both DELETE endpoints are rate-limited to 500/min, so `async_delete_with_retry` retries on 429, 5xx, and network errors (real 4xx like 404-already-gone still stop), with no redundant sleep after the final attempt. It logs at error level if it ultimately gives up, so a persistent leak is visible rather than silent.

## Verification

- `cargo fmt`, `cargo clippy --features soniox -- -D warnings`, and default `cargo clippy -- -D warnings`: all clean
- `cargo test --features soniox`: 41 passed, 0 failed

## Notes

No automated test for the cleanup itself — it's all live HTTP with no reqwest mocking elsewhere in the codebase. Deferred (judgment calls): detaching success-path cleanup via `tokio::spawn` to remove ~6s latency under sustained rate-limiting, and honoring the `Retry-After` header instead of fixed backoff.

### DIFF
--- a/src/transcribe/soniox.rs
+++ b/src/transcribe/soniox.rs
@@ -798,110 +798,123 @@ impl SonioxTranscriber {
         let job_id = job.id;
         tracing::debug!("Soniox async: created job_id={}", job_id);
 
-        // 3. Poll until completed or error.
-        let poll_interval = ASYNC_POLL_INTERVAL;
-        let max_wait = Duration::from_secs(self.config.async_max_wait_secs.max(5));
-        let poll_start = std::time::Instant::now();
-        loop {
-            if poll_start.elapsed() > max_wait {
-                self.async_cleanup(client, &auth, &job_id, Some(&file_id))
-                    .await;
-                return Err(TranscribeError::InferenceFailed(format!(
-                    "Soniox async: job {} did not complete within {}s",
-                    job_id,
-                    max_wait.as_secs()
-                )));
+        // 3-4. Poll until completed, then fetch + parse the transcript. Wrapped
+        // in an async block so EVERY exit (including `?` on transport/parse
+        // errors) falls through to the single cleanup below; otherwise a
+        // transient poll/fetch error returns early and leaks both the
+        // transcription and its uploaded file.
+        let result: Result<TranscriptResponse, TranscribeError> = async {
+            let poll_interval = ASYNC_POLL_INTERVAL;
+            let max_wait = Duration::from_secs(self.config.async_max_wait_secs.max(5));
+            let poll_start = std::time::Instant::now();
+            loop {
+                if poll_start.elapsed() > max_wait {
+                    return Err(TranscribeError::InferenceFailed(format!(
+                        "Soniox async: job {} did not complete within {}s",
+                        job_id,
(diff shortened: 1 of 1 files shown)


seapy:
Independently hit this exact bug and can confirm both the diagnosis and the fix.

**Incident data** (v0.7.5 `8d49248`, meeting mode on the Soniox async backend, `chunk_duration_secs = 10`): after ~2 days of meetings, every upload started failing org-wide with `429 limit_exceeded` — "Total file count limit has been exceeded for your organization. Please delete some." — and the live transcript silently stopped mid-meeting. Account state at that moment matched this PR's diagnosis precisely:

- transcriptions: **0** — `DELETE /v1/transcriptions/{id}` was succeeding on every chunk
- files: **exactly 1,000** `voxtype.wav` uploads remaining — the org file cap

So the API-reference wording ("Permanently deletes a transcription and its associated files") verifiably does not hold for the uploaded audio file: it has to be removed via `DELETE /v1/files/{id}` separately, exactly as this PR does (and as Soniox's own async examples do).

**Fix verification**: cherry-picked the three commits here onto `8d49248` and rebuilt with `--features soniox,osd-gtk4,gpu-vulkan`. `cargo test --lib`: 777 passed, 1 failed — the failure is `setup::parakeet::tests::detect_available_backends_returns_vec` (environment-dependent backend-count assertion), unrelated to this PR. Live meeting test: the account file count went N → N+1 while a chunk was in flight and returned to N after each chunk completed, staying flat after `meeting stop` — where the unpatched build leaked exactly one file per chunk, monotonically.

One note for reviewers: at 10-second chunks the cap is reached within roughly two days of normal meeting use, and the failure mode is silent transcript loss mid-meeting — would be great to see this land.

materemias:
Following up now that v1.0.0 is released. CI is green and the file cleanup fix is still relevant. Could you review it for v1.0.1?

PR #515, opened 2026-06-29
fix(flake): enable Cohere in ONNX packages

repparw:
## Summary
- add `cohere` to the ONNX CPU feature set
- add `cohere-cuda` to the ONNX CUDA feature set
- leave MIGraphX unchanged because there is no `cohere-migraphx` feature and the source notes Cohere is not currently supported there

## Test
- `nix flake show --json --no-write-lock-file`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,6 +55,7 @@
           "paraformer"
           "dolphin"
           "omnilingual"
+          "cohere"
         ];
 
         onnxCudaFeatures = [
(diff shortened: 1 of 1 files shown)


repparw:
This came up while using the flake package with `programs.voxtype.settings.engine = "cohere"`: the ONNX package exposes the right runtime shape, but the binary is missing the `cohere` Cargo feature, so Cohere fails at startup. Adding these features makes the flake ONNX packages match the documented ONNX engine set.

DuskyElf (approved):
makes sense

PR #534, opened 2026-07-11
feat(output): add portal (RemoteDesktop) typing driver

### LINKED ISSUE #680: [Feature] Wayland Portal support
## Problem

On KDE Plasma 6 and GNOME there is no good keystroke driver in the default chain.

- `wtype` needs the `virtual-keyboard` Wayland protocol, which neither compositor implements, so it fails outright.
- `dotool` and `ydotool` go through uinput, which means adding the user to the `input` group, or running a daemon, or both.
- `eitype` does work through libei, but it registers a fresh session on every call. The docs in this repo already note that this flickers the tray indicator and adds latency during streaming dictation.
- That leaves the clipboard fallback, which overwrites whatever the user had on their clipboard.

So the out-of-the-box experience on the two most common desktops is either a permissions detour or a driver that visibly misbehaves.

## Proposed Solution

An output driver that types through the XDG RemoteDesktop portal (`org.freedesktop.portal.RemoteDesktop`), holding one session for the life of the daemon and injecting keysyms with `NotifyKeyboardKeysym`.

- Works on KDE, GNOME and wlroots compositors, since they all ship the portal.
- No uinput, no `input` group, no extra daemon.
- One session per daemon rather than one per keystroke batch, so no tray flicker and no per-call setup latency.
- Layout-independent: it sends codepoints directly rather than going through XKB, so no `DOTOOL_XKB_LAYOUT` style configuration.
- The portal issues a restore token, so the consent dialog appears once rather than on every daemon start.

(Note that without a `.desktop` file the portal has no app id, so the consent dialog says "some application" rather than "Voxtype". Shipping one would fix the wording for `eitype` too.)

The question is whether this should be on by default. Putting the portal first would give any Wayland users a working default without configuring anything, at the cost of a one-time consent dialog for everyone. I do not mind either way and would rather have your call before it ships.

## Alternatives Considered

- **`eitype`**: already works on these desktops, but the per-call session registration is exactly what makes it unpleasant during streaming dictation.
- **uinput drivers**: `dotool` and `ydotool` work everywhere but require the `input` group or a running daemon, which is the friction this is trying to remove.
- **Clipboard fallback**: functional, but it clobbers the clipboard and needs a paste step.

## Additional Context

KeePassXC took the same approach for autotype on Wayland: https://github.com/keepassxreboot/keepassxc/pull/13359

There is a working implementation in #534, tested on KDE Plasma 6.7 (Wayland), Arch Linux.

Hoeze:
## Description

Adds a `portal` output driver that types through the XDG RemoteDesktop portal, holding one persistent session for the life of the daemon and injecting Unicode keysyms via NotifyKeyboardKeysym.

On KDE Plasma 6 and GNOME, `wtype` does not work (no virtual-keyboard protocol), and `eitype` re-registers a fresh libei session on every call, which this repo's own docs note flickers the tray indicator and adds latency during streaming dictation. The portal driver registers once, so it avoids both, needs no uinput / `input` group, and is layout-independent (sends codepoints directly, ignoring the XKB layout hints).
I followed the logic of the [KeePassXC autotype implementation](https://github.com/keepassxreboot/keepassxc/pull/13359).

- New `src/output/portal.rs`, wired into the driver enum, factory, and error type.
- Speaks the portal directly over zbus (no wrapper crate). The MPRIS code in `media.rs` uses only the stable Connection/Proxy core and migrated unchanged.
- Ships `packaging/voxtype.desktop` so the portal names consent dialogs "Voxtype" instead of "some application" (also fixes this for `eitype`).
- `voxtype setup` reports the driver, including a probe that checks portal availability without triggering a consent dialog.
- Enables the `portal` feature by default and puts it first in the default chain.

**Behavior change, please review.** On by default, portal is tried first and supersedes `wtype` where the portal is available, and users get a one-time consent dialog on first dictation. That changes current default behavior.

## Related Issue

Depends on #679, which needs to merge first.

Fixes #680.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

Nothing breaks: no config migration, no removed option, and the chain still falls through to `wtype` where the portal is unavailable.
The one commit that changes existing behavior is `feat(output): enable portal by default and make it the primary driver`, described above.

## Testing

- [x] I have run `cargo clippy -- -D warnings` with no warnings
- [x] I have run `cargo fmt`

I manually verified typing + consent dialog on KDE Plasma 6.7 Wayland. 1021 tests pass, and clippy is clean both with the portal feature on and with `--no-default-features`, so the portal-off build is not broken.

## Documentation

- [x] I have updated documentation as needed
- [ ] No documentation changes are needed

`docs/CONFIGURATION.md` and `docs/USER_MANUAL.md` list the driver in the fallback chain.

## Pre-merge checklist

- [ ] Wait for `ci-success` to pass before marking Ready for Review (draft mode is encouraged while CI is red)

## Additional Notes

The portal caches its restore token under `$XDG_CACHE_HOME/voxtype/`, so the consent dialog appears once rather than on every daemon start.

Please note that large parts of this PR are AI-generated. I reviewed the changes, but I do not understand Rust well enough to judge whether src/output/portal.rs is a sane implementation or overly complicated.


Hoeze:
Dear @peteonrails, may I ask you considering this PR and approve CI?

Hoeze:
Hey @peteonrails, could you please have a check of this feature? Would be nice to have it in a released version :)

PR #554, opened 2026-08-03
macOS: keep the menu bar app out of the Dock

ivancis1:
## Description

The macOS menu bar helper shows a Dock icon and appears in Cmd-Tab, even though `LSUIElement` is set in the app bundle's `Info.plist` and the process only owns a status bar item (no window).

Root cause: `src/menubar.rs` builds the tray's event loop with `EventLoopBuilder::new().build()`. tao's own docs state the activation policy "is set to `NSApplicationActivationPolicyRegular` by default," and it applies that policy when the event loop starts, which overrides `LSUIElement` from the bundle. No Info.plist change can fix this; it has to be set on the event loop itself.

Fix: call `EventLoopExtMacOS::set_activation_policy(ActivationPolicy::Accessory)` before `run()`.

Verified with `lsappinfo info -only ApplicationType <ASN>`:
- Before: `ApplicationType=Foreground`
- After: `ApplicationType=UIElement`

Also confirmed the Dock icon and Cmd-Tab entry are gone after installing via `voxtype setup app-bundle`, and that the menu bar item, hotkey, and recording still work normally.

## Related Issue

No existing issue; found this while setting up the menu bar app fresh.

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## Testing

- [x] I have run `cargo clippy -- -D warnings` with no warnings introduced by this change (the crate has 12 pre-existing clippy errors on `dev`, none touched by this diff; `cargo clippy --all-targets --no-deps --features gpu-metal -- -D warnings` reports the same 12 with or without this branch applied)
- [x] I have run `cargo fmt`

## Documentation

- [ ] I have updated documentation as needed
- [x] No documentation changes are needed

## Pre-merge checklist

- [x] Wait for `ci-success` to pass before marking Ready for Review (draft mode is encouraged while CI is red)

## Additional Notes

This is a 2-line fix isolated to `src/menubar.rs`, built and tested on macOS 26.5.2 (Apple Silicon) with `--features gpu-metal`. Opened as a companion to #<SF_SYMBOLS_PR> (native SF Symbols menu bar icon), kept separate since they're independent changes to the same file.

### DIFF
--- a/src/menubar.rs
+++ b/src/menubar.rs
@@ -11,6 +11,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tao::event_loop::{ControlFlow, EventLoopBuilder};
+use tao::platform::macos::{ActivationPolicy, EventLoopExtMacOS};
 use tray_icon::{
     menu::{CheckMenuItem, Menu, MenuEvent, MenuItem, PredefinedMenuItem, Submenu},
     TrayIconBuilder,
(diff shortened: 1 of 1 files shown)


vantuan5644:
Confirmed on 1.0.1 (Apple Silicon): a stock `Voxtype.app` reports `ApplicationType=Foreground` despite `LSUIElement=true` in the bundle's Info.plist, exactly as described. In the shipped binary the `setActivationPolicy:` argument is loaded from tao's `AuxDelegateState.activation_policy`, so the bundle key never gets a say.

On the CI block: the only red job is `clippy`, and its single error is `useless use of format!` at `src/tui/compositor_bindings.rs:441`, which is unrelated to this diff. `dev` has since fixed that line in `71db5a0` ("Drop a format! that clippy now rejects"). This diff still applies cleanly to `dev` at `e325106`, so a rebase should be enough to turn `ci-success` green.

Worth reviving; it's the correct fix and no Info.plist change can substitute for it.

PR #568, opened 2026-08-16
Stop reporting GPU backends as active when they cannot load (closes #444)

### LINKED ISSUE #444: Arch package missing 'migraphx' dependency — voxtype-onnx-migraphx fails to load provider library
## Summary

The Arch Linux `voxtype` package's `Depends On` list does not include `migraphx`, even though the package ships the `voxtype-onnx-migraphx` binary that requires it. On a fresh install with AMD GPU + ROCm runtime, MIGraphX GPU acceleration cannot work because `libmigraphx_c.so.3` (provided by the `migraphx` package, in the `extra` repo) isn't pulled in.

The daemon does not crash — it falls back to CPU silently. So a user who clicks "Enable GPU" and sees `voxtype setup gpu --status` report `Active backend: ONNX GPU (MIGraphX)` has every reason to think GPU acceleration is working, when it actually isn't.

## Reproduction

Fresh Arch install with AMD GPU and ROCm:

```bash
pacman -Qi voxtype | grep "Depends On"
# Depends On      : alsa-lib  curl  gcc-libs  glibc
# (no migraphx, no rocm-hip-runtime)

pacman -Ss '^migraphx$'
# extra/migraphx 7.2.3-1
#     AMD's graph optimization engine
# (NOT installed by default)

sudo voxtype setup gpu --enable
systemctl --user restart voxtype
journalctl --user -u voxtype -g "MIGraphX"
```

Error in journal:
```
ERROR An error occurred when attempting to register `MIGraphXExecutionProvider`:
  [ONNXRuntimeError] : 1 : FAIL : Failed to load library
  /usr/lib/voxtype/migraphx/libonnxruntime_providers_migraphx.so
  with error: libmigraphx_c.so.3: cannot open shared object file: No such file or directory
```

After `sudo pacman -S migraphx`, the provider registers and GPU inference works.

## Suggested fixes

The Arch `voxtype` package's PKGBUILD should declare:

```bash
optdepends=(
    'migraphx: AMD GPU acceleration via MIGraphX execution provider'
    'rocm-hip-runtime: required by migraphx'
    # ...existing optdepends
)
```

or treat it as a hard dependency when the `voxtype-onnx-migraphx` variant is shipped. (`optdepends` is probably cleaner since the package also ships CUDA and Vulkan variants that don't need MIGraphX.)

Additionally, `voxtype setup gpu --status` should distinguish "binary symlinked to MIGraphX variant" from "MIGraphX provider successfully loaded at runtime" — the former is currently reported as "Active backend: ONNX GPU (MIGraphX)" even when the actual provider failed to register. A simple fix: check `journalctl` for the most recent registration result or expose a `--probe` flag that does a load-only check.

## Environment

- VoxType 0.7.3, arch `voxtype` package
- AMD Radeon RX 7900 XT, ROCm 7.2.3 (`hsa-rocr`, `rocm-hip-runtime`, etc. all installed)
- `migraphx` package was NOT installed despite ROCm being present
- After `pacman -S migraphx`, provider loads successfully

Related: voxtype#443 (path-lookup wrapper).

Reported via voxtype-tray.

peteonrails:
Fixes #444. Wanted before the Omarchy Quattro compatibility testing, because a status lie makes a green test result meaningless.

## The lie

`setup gpu --status` built its answer entirely from the symlink target's filename:

```rust
"voxtype-onnx-migraphx" => "ONNX GPU (MIGraphX)",
```

Nothing verified the execution provider could load. The GPU binaries dlopen `libonnxruntime_providers_<ep>.so` from their own directory; when the system libraries that provider links against are missing, ORT fails to register the EP and falls back to CPU without surfacing anything. The user is told they have GPU acceleration and measures CPU speed.

## The check

Resolve the provider library next to the real binary, run `ldd`, and report any `=> not found` dependencies. On a healthy AMD host every dependency resolves through `/opt/rocm`:

```
libmigraphx_c.so.3 => /opt/rocm/lib/libmigraphx_c.so.3
libamdhip64.so.7 => /opt/rocm/lib/libamdhip64.so.7
```

On the reporter's host those same lines read `not found`, which is precisely the state that used to print "active". Output becomes:

```
Active backend: ONNX GPU (MIGraphX)
  Binary: /usr/lib/voxtype/voxtype-onnx-migraphx

  WARNING: this backend is selected but cannot load. ONNX Runtime will
           fall back to CPU. Missing shared libraries:
             libamdhip64.so.7
             libmigraphx_c.so.3
           Install the AMD runtime, e.g. on Arch:
             sudo pacman -S migraphx rocm-hip-runtime
```

The same check runs on `--enable`, since selecting an unloadable backend is where the false belief starts rather than where it gets discovered.

## Packaging half

`packaging/arch-bin-rc/PKGBUILD` listed `rocm-hip-runtime` but not `migraphx`. The provider needs `libmigraphx_c.so.3`, which ships in the `migraphx` package, so the optdepend was incomplete. Added it.

Left as optdepends rather than depends: making ROCm a hard dependency would pull it onto every Arch install including CPU-only ones.

Note `packaging/arch-bin/` (the stable channel) is a gitignored nested AUR repo, so its PKGBUILD needs the same edit applied out-of-band.

## Verification

- 4 unit tests on the `ldd` parser: unresolved deps, all-resolved, lines without `=>` (vdso, "statically linked"), and dedupe
- `cargo clippy --all-targets -- -D warnings` clean, `cargo test` 802 passing
- Confirmed against the real provider libraries in `/usr/lib/voxtype/migraphx/` that a healthy host produces no `not found` lines

Not verified end to end: reproducing the failing case needs a host without the AMD runtime, and this machine has ROCm installed and runs Whisper CPU, so the ONNX status branch was not reachable without modifying the install. The parser is covered by tests against captured real output; the Quattro VM is the natural place to exercise the whole path.

### DIFF
--- a/packaging/arch-bin-rc/PKGBUILD
+++ b/packaging/arch-bin-rc/PKGBUILD
@@ -41,6 +41,7 @@ optdepends=(
     'cuda: GPU acceleration via CUDA 13 for ONNX engines (NVIDIA GPUs, requires driver 580+)'
     'cuda12.6: GPU acceleration via CUDA 12 for ONNX engines (older NVIDIA setups)'
     'rocm-hip-runtime: GPU acceleration via MIGraphX for ONNX engines (AMD GPUs)'
+    'migraphx: required alongside rocm-hip-runtime for the MIGraphX provider (#444)'
     'ollama: local AI summarization for meeting mode'
     'gtk4-layer-shell: runtime for the GTK4 on-screen mic visualizer (voxtype-osd-gtk4)'
     'quickshell: Quickshell-based on-screen display frontend (opt in via [osd] frontend = "quickshell")'
--- a/src/setup/gpu.rs
+++ b/src/setup/gpu.rs
@@ -17,7 +17,7 @@
 //!
 //! This sets VK_LOADER_DRIVERS_SELECT internally to filter Vulkan ICDs.
 
-use super::binary::install_active_binary;
+use super::binary::{install_active_binary, resolve_active_binary};
 use std::fs;
 use std::os::unix::fs::symlink;
 use std::path::Path;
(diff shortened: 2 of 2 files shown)


peteonrails:
Rebased onto current `dev` and added a second commit after testing against a real MIGraphX install. The original fix was incomplete.

## The status was wrong in a second, larger way

On a machine where `/usr/bin/voxtype` is the dispatch wrapper (the layout every GPU/ONNX install uses) and `engine = "parakeet"`:

```
$ voxtype setup gpu --status
Active backend: CPU (native)
  Binary: /usr/lib/voxtype/voxtype-native

$ voxtype setup onnx --status
Active engine: Parakeet
  Backend: ONNX (MIGraphX)
```

Two status commands, opposite answers, same machine. `fs::canonicalize` on the wrapper returns the wrapper, whose filename is `voxtype`, so every `contains("onnx")` check concluded no ONNX backend was active and the whole status fell through to the Whisper branch.

That also made the first commit's warning **unreachable on exactly the configuration it was written for**, since the provider check only runs inside the parakeet branch.

`setup onnx --status` already got this right via `binary::resolve_active_binary`, which handles symlink and wrapper alike. Now shared instead of duplicated.

## Verified on real hardware, not stubs

An AMD host with ROCm installed and the full ONNX variant set in `/usr/lib/voxtype/`:

- `setup gpu --status` now reports `ONNX GPU (MIGraphX)` and agrees with `setup onnx --status`
- No warning in the healthy state; `ldd` on the MIGraphX provider shows 0 unresolved deps
- Pointing the wrapper at `cuda-13`, whose CUDA libraries are absent on this box, produces:

```
Active backend: ONNX GPU (CUDA 13)
  Binary: /usr/lib/voxtype/voxtype-onnx-cuda-13

  WARNING: this backend is selected but cannot load. ONNX Runtime will
           fall back to CPU. Missing shared libraries:
             libcublas.so.13
             libcublasLt.so.13
             libcudart.so.13
             libcudnn.so.9
             libcufft.so.12
             libcurand.so.10
           Install the matching NVIDIA CUDA runtime for this variant.
```

- Wrapper restored, checksum verified identical, daemon still active, no false positive

`cargo clippy --all-targets -- -D warnings` clean, 802 tests passing.

PR #569, opened 2026-08-16
docs: troubleshoot IBus reordering of non-ASCII input on GNOME

lunetics:
**Cause:** On GNOME, IBus reorders non-ASCII key events relative to ASCII when synthetic input arrives faster than human typing. Dictated umlauts/accents end up clustered at the front of the output in GTK apps while the voxtype log shows a correct transcription. The corruption is silent, looks like a voxtype or eitype bug, and the troubleshooting guide currently recommends eitype for GNOME without mentioning it.

**Impact:** New troubleshooting section under Output Problems: symptom (with the log-vs-output discriminator), cause with upstream references (ibus/ibus#2934, Adam-D-Lewis/eitype#21), paste mode as the reliable solution, `type_delay_ms` and `GTK_IM_MODULE=simple` as partial alternatives. Docs only, no code changes.

**Verification:** All claims measured on Ubuntu 24.04 / GNOME 46 / IBus 1.5.29-rc2: fixed-string eitype runs into `cat`, receiver comparison (GNOME Terminal and GNOME Text Editor scrambled; alacritty and Chrome correct; same GTK app correct with `GTK_IM_MODULE=simple`), delay sweep (25 ms shrinks the displacement to 2-3 positions, ~100 ms/key would be needed). Details in the two linked issues.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -843,6 +843,36 @@ You can also enable it via CLI flag (`--wtype-shift-prefix`) or environment vari
 type_delay_ms = 10  # Try 10-50ms
 ```
 
+### Non-ASCII characters move to the front of the text (GNOME, type mode)
+
+**Symptom:** Dictating text with non-ASCII characters (umlauts, accents, ß)
+into GTK applications (GNOME Terminal, GNOME Text Editor) delivers them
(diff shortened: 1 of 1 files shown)


peteonrails (approved):

peteonrails:
@lunetics Thank you!

PR #584, opened 2026-08-20
Resume paused media when a dictation is cancelled

lunetics:
**Cause:** Three of the four cancel paths in `Daemon::run()` return to `Idle` without undoing what recording start applied to playback: paused MPRIS players stay paused, ducked streams stay quiet. Only `cancel_streaming_to_idle()` restores both.

**Impact:** With `pause_media = true`, cancelling a dictation leaves music silent until some later completed dictation happens to resume it. The fix adds the two existing restore calls to each path, in the order the streaming path already uses.

No test (the cancel arms sit inside the `tokio::select!` loop, no seam); not compiled locally.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -3131,6 +3131,8 @@ impl Daemon {
                                 cleanup_model_override();
                                 cleanup_profile_override();
                                 cleanup_bool_override("smart_auto_submit");
+                                self.restore_ducked_media_streams();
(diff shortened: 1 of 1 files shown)


peteonrails:
Thanks for finding this!

#561 fixed it already: `restore_recording_media()` covers all three cancel paths now. Closing as already fixed.

Appreciate you opening the PR!

PR #585, opened 2026-08-20
Fade media ducking in and out instead of jumping

lunetics:
**Why:** ducking sets the volume in a single `pactl` call, so music snaps down at recording start and snaps back at the end. The jump is audible — exactly the attention grab ducking is meant to avoid.

**Changes:** a ramp walks the volume to the target in 20 ms steps, same duration down and up. `duck_media_fade_ms` defaults to 150 ms; `0` keeps the old instant behaviour. Only intermediate steps are ramped — the exact target and restore values are still set by the existing calls, so their error handling and the stored originals are untouched.

**Verification:** `cargo test --lib audio::media` → 7 passed (4 new: fade shorter than a step is a no-op, ramp down, ramp up, monotonic and inside the endpoints). `cargo clippy --lib` clean. Built and run locally as the vulkan variant on GNOME/Wayland; ducking fades audibly in both directions.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -410,6 +410,24 @@ applied to each stream's current per-channel volume, not to a fixed 100% base:
 original per-channel volumes are restored when recording stops. Values above
 `150` are clamped by the CLI override.
 
+### duck_media_fade_ms
+
+**Type:** Integer
+**Default:** `150`
+**Required:** No
+
+Fade duration in milliseconds for the ducking ramp. The same duration is used
+going down at recording start and coming back up when recording ends, so media
+slides out of the way instead of jumping. `0` restores the previous instant
+behaviour. Durations shorter than one 20 ms step are treated as `0`.
+
+```toml
+[audio]
+duck_media = true
+duck_media_volume_percent = 70
+duck_media_fade_ms = 150
+```
+
 ---
 
 ## [audio.feedback]
--- a/src/audio/media.rs
+++ b/src/audio/media.rs
@@ -120,7 +120,61 @@ mod imp {
     /// enabled alongside `pause_media`, but when both are enabled the pause
     /// feature keeps its existing start/stop behavior and ducking only manages
     /// stream volume.
-    pub async fn duck_playing_audio(volume_percent: u8) -> Vec<DuckedMediaStream> {
+    /// Interval between ramp steps. Short enough to sound continuous, long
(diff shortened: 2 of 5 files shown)


peteonrails:
@lunetics Thanks for this! I like the fade: I think it is the right idea. A few things to resolve:

1) The down-fade delays the microphone open event. duck_playing_audio is awaited in suppress_recording_media(), which runs before start_recording_capture(). With the 150 ms default that's 7 steps of pactl-per-stream plus 20 ms sleeps before the mic opens. Push-to-talk users start talking at keypress, so this can clip the first syllables. In my testing, I sawmy first word get dropped or mangled. 

I can think of two ways to address this: default to 0 so the fade is opt-in, or run the down-ramp concurrently with capture instead of blocking it. I think I prefer the concurrent version, what do you think? It's a bit of a tradeoff since media during the fade is initially not quieted.

2) There is a restore race condition: when restore_ducked_audio has already been spawned, then you start a new recording inside that window and the new duck captures mid-fade volumes as its originals, while the old restore task finishes by writing the old originals back, un-ducking media during the new recording. I think the way abound this is to keep the JoinHandle and abort or await it before ducking again.

A couple smaller nits:

- The rustdoc got shuffled in the edit: duck_playing_audio's old doc comment now sits on FADE_STEP_MS, and ramp_factors carries two stacked doc blocks, one of which describes ramp_streams. 

- duck_media_fade_ms needs a CLI flag next to --duck-media and --duck-media-volume, so scripts and keybindings can override it without editing the config file. Every option should get all three: CLI, env var, config.

Would you rather I take the concurrent-ramp change? Happy to push it to your branch, and to merge once the mic delay is sorted out.

peteonrails:
Closing in favour of #659, which carries this commit and finishes the three items from my review rather than asking you for another round.

For the record on each: the down-fade now runs on a spawned task with the originals captured before it starts, so the microphone opens immediately — that was the one I'd actually reproduced, first words getting clipped. The restore race turned out to be worse than I described: because restore runs for the whole fade now, a second recording inside that window captured mid-fade volumes as the new originals, so every fast toggle ratcheted media permanently quieter. The daemon now holds the fade task so duck and restore serialize against each other. And `--duck-media-fade-ms` plus the env var and schema entry are in, so `config set audio.duck_media_fade_ms` works.

The fade was the right idea and it sounds good. Thanks for pushing it.

PR #586, opened 2026-08-20
Make duck_media_volume_percent mean what it says

lunetics:
Builds on #585 (its commit is included in this branch; merging that first shrinks this diff to one commit).

**Root Cause:** the ducking target scales PulseAudio's volume percentage directly, but that scale is cubic — the amplitude is `(percent/100)^3`. The configured reduction is therefore cubed: `50` leaves 12.5% of the amplitude (-18 dB) instead of half, `30` is effectively mute, and `CONFIGURATION.md`'s own example ("`50` keeps it at half") doesn't match the behaviour.

**The Fix:** the value now means the fraction of its current amplitude a stream keeps. `scaled_volumes()` compensates the cubic scale with a cube root, so `50` halves the amplitude (-6 dB); docs, config template and CLI help all describe that. The fade ramp (#585) interpolates in the amplitude domain and gets perceptually smoother as a side effect.

**BC note:** behaviour changes for configs tuned against v1.0.0-rc builds — the same number now ducks audibly less. The option has never shipped in a stable release, so pre-1.0 seemed like the right moment to fix the meaning rather than document the surprise; the docs carry a migration note. Happy to convert this to a new field with a deprecation path instead if you'd rather keep rc-to-rc compatibility.

**Validation:** `cargo test --lib audio::media` → 10 passed, including `half_amplitude_is_six_db_not_eighteen` (fails without the fix: the old code returns `50%` where half the amplitude requires `79%`). `cargo clippy --lib` clean.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -410,6 +410,24 @@ applied to each stream's current per-channel volume, not to a fixed 100% base:
 original per-channel volumes are restored when recording stops. Values above
 `150` are clamped by the CLI override.
 
+### duck_media_fade_ms
+
+**Type:** Integer
+**Default:** `150`
+**Required:** No
+
+Fade duration in milliseconds for the ducking ramp. The same duration is used
+going down at recording start and coming back up when recording ends, so media
+slides out of the way instead of jumping. `0` restores the previous instant
+behaviour. Durations shorter than one 20 ms step are treated as `0`.
+
+```toml
+[audio]
+duck_media = true
+duck_media_volume_percent = 70
+duck_media_fade_ms = 150
+```
+
 ---
 
 ## [audio.feedback]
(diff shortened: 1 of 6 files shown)


lunetics:
The clippy failure is not from this change: it fires in `src/tui/compositor_bindings.rs:441` (`useless_format` in the Niri stop binding), a file this PR does not touch — it surfaces on any cold clippy run against current `dev`. Fix filed as #587; once that lands, a re-run here should go green.

peteonrails:
I think you've got the right idea.

Rather than send you round again on the #585 review items, I took your commits and finished them on top: #659. Your commits keep their authorship and you're co-author on mine.

I did make one call differently from what you offered. Correcting the meaning without moving the default would have made ducking much gentler for everyone, because the shipped 70 was going through the cubic scale and actually leaving 0.70³ = 0.343 of the amplitude. That's -9.3 dB today. As a plain amplitude it becomes -3.1 dB. So I moved the default to 34, which keeps the depth people already have, and put a test on it so it doesn't get rounded back to 70 later.

You offered a new field with a deprecation path instead. Do you think that's the better road? It's the one decision in #659 I'd want a second opinion on.

Closing this in favour of #659.

PR #587, opened 2026-08-20
Fix clippy useless_format in the Niri stop binding

lunetics:
**Cause:** the Niri `Role::Stop` arm wraps a plain string literal in `format!()` with no interpolation. CI runs `cargo clippy -- -D warnings`, so any pull request that gets a cold clippy run currently fails on this line regardless of its own changes (first seen on the CI run for #586: `src/tui/compositor_bindings.rs:441`, `-D clippy::useless-format`).

**Impact:** unblocks clippy for every open and future PR against `dev`. The literal becomes a `.to_string()` call; the generated binding text is byte-identical.


peteonrails:
Already fixed on dev — `src/tui/compositor_bindings.rs:442` uses `.to_string()` now, which is why this shows as conflicting. Closing.

Thanks for catching it either way.

PR #589, opened 2026-08-21
fix(hotkey): honor hotkey release when modifier is released first

iacore:
**Bug:** With push-to-talk and a modifier chord (e.g. `Meta+V`), lifting the chord commonly delivers the modifier's release event before the hotkey key's. The evdev listener gated **both** press and release on `modifiers_satisfied`, so in that order the hotkey key's release was dropped: `is_pressed` stayed true, recording never stopped (until the `max_duration_secs` cap), the OSD stayed open, and the eventual transcription of the tail produced empty text that was pasted as nothing.

Reported as: voxtype OSD "never closes" after dictation; the transcription arrives empty (in a TUI app, shown as "Clipboard is empty").

**Fix:** Extract the press/release classification into a pure `hotkey_event()` state machine:
- Presses still require the required modifiers held (`1 if !is_pressed && modifiers_satisfied`).
- Releases now always disengage an engaged hotkey (`0 if is_pressed`), regardless of current modifier state — a released hotkey key ends the recording.

**Tests:** Added unit tests for the state machine: press-with-modifiers, press-without-modifiers, repeat, release-with-modifiers, release-without-modifiers (the fix), and stray release. Full lib suite passes (847 tests).

**Repro:** push-to-talk with `modifiers = ["LEFTMETA"]`, press Meta+V, then release Meta before V. The release event is dropped; recording continues until the cap.

### DIFF
--- a/src/hotkey/evdev_listener.rs
+++ b/src/hotkey/evdev_listener.rs
@@ -440,6 +440,31 @@ fn reset_for_device_change(
     was_pressed.then_some(HotkeyEvent::Released)
 }
 
+/// Hotkey state transition for a target-key event.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum HotkeyTransition {
+    /// The hotkey key was pressed with the required modifiers held.
+    Pressed,
+    /// The hotkey key was released while the hotkey was engaged.
+    Released,
+}
+
+/// Classify a target-key event as a hotkey press, release, or neither.
+///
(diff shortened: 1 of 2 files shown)


peteonrails:
@iacore Been chasing modifier key relkease issues for a loooooong time, so if this solves it you get many kudos. I have a couple of requests: 1) would you mind submitting signed commits and 2) could you check the failing clippy / fmt issues?

Issue #283, opened 2026-03-27
[Feature] Having streaming for text maybe?


AceCodePt:
## Problem

I want to see the text live as I speak it.

## Proposed Solution

Pretty self explanetory.

## Alternatives Considered

No.

## Additional Context

No. 
But this project is awesome. I want to attach it to opencode.

peteonrails:
Experimental streaming support is in 0.7.2, which is building now.

PR #282, opened 2026-03-26
[EXPERIMENT] Add OpenVINO GenAI engine backend for NPU/GPU acceleration

TsaiGaggery:
## Summary

This adds a new `openvino-genai` transcription engine that uses Intel OpenVINO GenAI's WhisperPipeline for hardware-accelerated speech-to-text on Intel NPU (AI Boost) and GPU. The implementation uses direct Rust FFI to `libopenvino_genai_c.so` with no Python or subprocess dependencies.

**Status: Experimental** — tested on Dell XPS 16 (Panther Lake) with OpenVINO 2026.0. Marked as draft for review and feedback.

## How It Works

```
Copilot key -> voxtype daemon -> PipeWire capture (16kHz f32)
  -> Rust FFI -> libopenvino_genai_c.so -> WhisperPipeline
  -> NPU inference -> transcribed text -> wtype (at cursor)
```

The `Transcriber` trait implementation is straightforward — `transcribe(&self, &[f32]) -> String` calls through FFI to the OpenVINO GenAI C API's `ov_genai_whisper_pipeline_generate()`.

## Changes

| File | Change |
|------|--------|
| `Cargo.toml` | Add `openvino-genai` feature flag |
| `src/transcribe/openvino_genai.rs` | **New**: FFI bindings + `OpenvinoGenaiTranscriber` |
| `src/config.rs` | Add `OpenvinoGenaiConfig`, `OpenvinoGenai` enum variant, `OPENVINO_DEVICE` env var |
| `src/transcribe/mod.rs` | Wire up engine factory with default config fallback |
| `src/daemon.rs` | Add `OpenvinoGenai` match arms |
| `src/vad/mod.rs` | Add `OpenvinoGenai` -> Energy VAD |
| `src/output/mod.rs` | Add emoji indicator |

## Usage

```bash
# Build
cargo build --release --features openvino-genai

# Run (CPU)
VOXTYPE_ENGINE=openvino voxtype daemon

# Run (NPU)
VOXTYPE_ENGINE=openvino OPENVINO_DEVICE=NPU voxtype daemon

# Config file
[openvino_genai]
model = "whisper-base-ov"
device = "NPU"
language = "en"
```

## Prerequisites

- OpenVINO 2026.0+ with GenAI (`pip install openvino openvino-genai`)
- `libopenvino_genai_c.so` (built from [openvino.genai](https://github.com/openvinotoolkit/openvino.genai) source)
[...]

PR #592, opened 2026-08-21
Sliding-window streaming transcription + OpenVINO NPU/GPU backend

tslove923:
## Description

Two things, sharing one engine:

1. **Sliding-window streaming transcription** for `[whisper] streaming = true` — ported from a companion project's `SlidingWindowTranscriber` (MIT), adapted to voxtype's `StreamingTranscriber` trait. Instead of segmenting on VAD boundaries, it re-transcribes a rolling audio buffer on an interval and emits only the stable-prefix delta, so text lands at the cursor incrementally as you speak instead of only after the hotkey releases.
2. **OpenVINO GenAI backend** (`engine = "openvino"`, feature-gated behind `--features openvino`) — runs Whisper via Intel's OpenVINO GenAI `WhisperPipeline` on NPU/GPU/CPU, falling back to CPU automatically if the requested device fails to initialize. The streaming engine above is backend-agnostic, so `[openvino] streaming = true` gets the same live-typing behavior on NPU hardware.

Full wiring: setup/model download flow (`voxtype setup model`), the `voxtype configure` TUI, daemon preload/streaming paths, VAD/notification/output-icon dispatch — the usual set of exhaustive `TranscriptionEngine` matches.

### Why NPU

The point isn't raw wall-clock speed — on `base.en`, NPU came out ~2.5x faster than CPU here (0.075s vs 0.185s mean per 4.75s clip, both already well past realtime since OpenVINO's CPU path is INT8-quantized, not a naive baseline). The real value is that inference moves off the P/E cores entirely and onto the NPU: a separate, always-on, low-power IP block on the SoC.

Measured with `getrusage` (process CPU-time, summed across all threads) around 10 back-to-back `transcribe()` calls on the same 4.75s clip:

| | CPU-seconds per call | Avg cores kept busy |
|---|---|---|
| CPU | 0.710s | 3.82 |
| NPU | 0.020s | 0.28 |

**~36x less CPU-core-time consumed on NPU.** CPU inference genuinely occupies ~3.8 cores for its duration; NPU inference barely touches the CPU (just data marshalling/queuing to the accelerator) while the actual compute runs on the separate IP block. That's the number that matters for laptop battery life and for not starving whatever else is running while you dictate — not the wall-clock latency, which undersells it since CPU inference is already fast in absolute terms by parallelizing across those 3.8 cores.

### GPU as a fallback on non-NPU hardware

For anyone without an NPU but with an Intel iGPU, the natural question is whether GPU is worth it over CPU. On `base.en`, GPU is only ~1.5x faster wall-clock than CPU — a much smaller win than NPU's, and it initially looked like a bad trade because of model-load time (see next section). With that fixed, GPU still pulls CPU-core-time down ~4x versus CPU (0.673s vs 2.915s per call, `getrusage`-measured the same way as the NPU table above), so it's a real if smaller offload win, and now viable with `on_demand_loading = true` as well since reload is fast (see below).

### Compiled model caching (NPU and GPU load time)

Both NPU and GPU pay a real cost before first inference: OpenVINO compiles the model graph into a device-specific blob (NPU: to NPU-ISA; GPU: OpenCL kernel compilation), and neither is fast — NPU measured 6-17s cold, GPU a few seconds, both on Meteor Lake with `base.en`. That's paid on *every* pipeline init, which for a long-running daemon means every restart.

`openvino-genai`'s `WhisperPipeline::new()` had no way to pass device properties to avoid this — unlike `LlmPipeline`/`VlmPipeline` in the same crate, which already expose `with_properties()` for exactly this reason. Opened [intel/openvino-rs#196](https://github.com/intel/openvino-rs/pull/196) upstream adding the same method for Whisper, and this PR pins a fork branch via `[patch.crates-io]` (with a comment pointing at the upstream PR, to drop once it lands) so `CACHE_DIR` gets wired through to a `$XDG_CACHE_HOME/voxtype/openvino` cache dir.

Measured on real hardware, cold vs. cached reload:

| Device | Cold | Cached | Speedup |
|---|---|---|---|
| NPU | 16.82s | 0.58-0.62s | ~28x |
| GPU | ~8s (true cold, no cache anywhere) | 0.50-0.58s | ~14x |
| CPU | 1.07s | 0.84-0.86s | ~1x (no meaningful compile step to cache) |

Practical effect: `on_demand_loading = true` (load only when recording starts, unload after) is now viable on NPU and GPU too, not just CPU — a cached reload is sub-second on either.

### Related work

I found these after starting, both still open/draft and tackling the same problem independently:
- #282 (`TsaiGaggery`) — landed on essentially the same design (direct Rust FFI to `libopenvino_genai_c.so`, no Python/subprocess). Good independent confirmation the approach is sound.
- #547 (`jacob-vincent-mink`) — broader scope, also adds HF download pathways for OpenVINO models.

Flagging in case a maintainer wants to consolidate. Two bugs below (in "Verified on hardware") were found through actual NPU testing and may be worth checking against either of those PRs if they hit the same failure.

## Verified on hardware

Tested end-to-end on real Intel NPU (Meteor Lake, `/dev/accel0`) — a real recording → transcribe → type cycle, then daily-driven through my own `voxtype.service` with `[openvino] streaming = true`. Screen recording of live streaming dictation attached below.

https://github.com/user-attachments/assets/d1aac3d3-b2f6-4349-b57a-149335d55c7d

Two real bugs surfaced during that testing and are fixed in this PR:

1. **`generate()` threw "unknown exception" on NPU** (worked on CPU/GPU). Root cause: we built the Whisper generation config via `WhisperGenerationConfig::new()` — a bare config disconnected from the loaded model, missing EOS/decoder-start token ids and the `is_multilingual` flag. The official `openvino_genai` C sample instead starts from `pipeline.get_generation_config()` (populated from the model's own `generation_config.json`) and only adjusts task/timestamps on top. Switched to that pattern. Also skip `set_language`/`set_task` entirely for English-only (`.en`) checkpoints — their vocabulary has no task/language tokens at all, so setting them looks up tokens that don't exist. This matches the failure shape of [openvinotoolkit/openvino.genai#4222](https://github.com/openvinotoolkit/openvino.genai/issues/4222).
2. **`Config::streaming_active()`** (which auto-promotes push-to-talk to toggle activation, because typing at the cursor while a key is physically held clobbers libinput's held-key tracking on Hyprland/Sway/River) had no `OpenVino` arm and silently fell through to `false`. Enabling `[openvino] streaming` with push-to-talk would have left recording permanently stuck open on the first real session. Fixed.

## Related Issue

Relates to #283 (streaming transcription request, closed) and the OpenVINO NPU interest in #282 / #547.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Testing

- [x] I have tested these changes locally (including on real NPU hardware, daily-driven through my own systemd service)
- [x] I have run `cargo test` and all tests pass (855+, with and without `--features openvino`)
- [x] I have run `cargo clippy -- -D warnings` with no warnings (both feature sets)
- [x] I have run `cargo fmt`

## Documentation

- [x] I have updated documentation as needed *(added a full `[openvino]` section to `docs/CONFIGURATION.md`, matching the existing per-engine sections)*

## Pre-merge checklist

- [ ] Wait for `ci-success` to pass before marking Ready for Review

## Additional Notes

`OpenVinoModelInfo`/model catalog covers `tiny.en`, `base.en`, `small.en`, `base`, `small` — downloaded via `huggingface-cli` from the `OpenVINO/whisper-*-int8-ov` org (not yet mirrored to the voxtype R2 CDN, same caveat as Cohere).

### Arch Linux setup (what it actually took to get NPU inference running)

Pacman `extra` gets you the core toolkit and device plugins:

```
pacman -S openvino openvino-intel-npu-plugin openvino-intel-gpu-plugin \
          intel-npu-driver level-zero-loader intel-compute-runtime
```

That's necessary but not sufficient. The Rust crate (`openvino-genai` / `openvino-genai-sys`, runtime-linked via `openvino-finder`) needs a separate **C API** library, `libopenvino_genai_c.so` — and neither the AUR `openvino-genai-bin` package nor the `openvino-genai` pip wheel ship it. Both only bundle the Python-binding `.so` (pybind11, a different linking mechanism entirely), which is a dead end for FFI from Rust/C/C++.

The actual fix: download Intel's official GenAI C++/C SDK archive directly —
`openvino_genai_ubuntu24_<version>_x86_64.tar.gz` from `storage.openvinotoolkit.org/repositories/openvino_genai/packages/<version>/linux/`, **version-matched to the installed core `openvino` package** (e.g. core `openvino` 2026.3.0 → SDK `openvino_genai_ubuntu24_2026.3.0.0_x86_64.tar.gz`) — extract it anywhere, and point `LD_LIBRARY_PATH` at `<extracted>/runtime/lib/intel64` (where `libopenvino_genai_c.so` actually lives) when running voxtype.

That same SDK archive's `samples/c/whisper_speech_recognition/whisper_speech_recognition.c` is what led to the `generate()` fix above — it's the canonical example of correct `WhisperPipeline` C API usage (`get_generation_config()` off the pipeline, not a bare `create()`), and there's no equivalent reference for this pattern in the Rust crate's own docs.

Worth a packaging note somewhere (`docs/INSTALL.md`, PKGBUILD `optdepends`) if this engine moves toward a real release story — as it stands, `openvino-genai-bin` on AUR being insufficient is a real trap for anyone trying this from the pacman package alone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3907,8 +3907,7 @@ dependencies = [
 [[package]]
 name = "openvino-finder"
 version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a527e1414732f5bcad520b9febf29fd92ef4bba100588b4591cba8370ab6cbda"
+source = "git+https://github.com/tslove923/openvino-rs?branch=whisper-pipeline-with-properties#8109638cfc78eb16953316627bf194d59ca521ab"
 dependencies = [
  "cfg-if",
  "log",
@@ -3917,8 +3916,7 @@ dependencies = [
 [[package]]
 name = "openvino-genai"
 version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58748dc7c7eb98f2706152eb9e3ce0f4023359e54a9607656d653f8b7af689a8"
+source = "git+https://github.com/tslove923/openvino-rs?branch=whisper-pipeline-with-properties#8109638cfc78eb16953316627bf194d59ca521ab"
 dependencies = [
  "openvino-finder",
  "openvino-genai-sys",
@@ -3927,8 +3925,7 @@ dependencies = [
 [[package]]
 name = "openvino-genai-sys"
 version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ce904847ba3b9328e412dd95741e432f75db3177e21742f80b22eb9dcef4cafe"
+source = "git+https://github.com/tslove923/openvino-rs?branch=whisper-pipeline-with-properties#8109638cfc78eb16953316627bf194d59ca521ab"
 dependencies = [
  "env_logger",
  "libloading 0.8.9",
@@ -3939,8 +3936,7 @@ dependencies = [
 [[package]]
 name = "openvino-sys"
 version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "637459f74d22a7bb772af9d3a5df61c88608d8daaa6fefe950849c7aabf891c3"
+source = "git+https://github.com/tslove923/openvino-rs?branch=whisper-pipeline-with-properties#8109638cfc78eb16953316627bf194d59ca521ab"
 dependencies = [
  "env_logger",
  "libloading 0.8.9",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,7 +84,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
- "anstyle-parse",
+ "anstyle-parse 0.2.7",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
+dependencies = [
+ "anstyle",
+ "anstyle-parse 1.0.0",
  "anstyle-query",
  "anstyle-wincon",
  "colorchoice",
@@ -103,6 +118,15 @@ name = "anstyle-parse"
 version = "0.2.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-parse"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
 dependencies = [
  "utf8parse",
 ]
@@ -652,7 +676,7 @@ version = "4.5.53"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d76b5d13eaa18c901fd2f7fca939fefe3a0727a953561fefdf3b2922b8569d00"
 dependencies = [
- "anstream",
+ "anstream 0.6.21",
  "anstyle",
  "clap_lex",
  "strsim",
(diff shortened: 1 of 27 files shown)


spencerbull:
@jacob-vincent-mink @peteonrails @tslove923 does this PR help us out so that we can include openvino-genai?

https://github.com/omacom-io/omarchy-pkgs/pull/189

I will work on the NPU driver in omarchy-pkgs as well, but can continue to use the above mentioned NPU driver until I get that included.

jacob-vincent-mink:
@spencerbull right deps identified. Likely need @peteonrails input for how they’d typically want to deliver these discrete capabilities.

e.g. install on blanket deps + iGPU on Intel system with no NPU, and so on. It also needs the full OV package, as mentioned, that includes the c-bind for the rust stuff to work (since the rust lib is a pure lib, does not carry the OV deps on purpose - Intel often treats OV as a global install). 

Not to hijack the PR, great feature! :)

I do believe the AUR stuff is out of date, but they may have updated the driver/package since last I looked.

tslove923:
Thanks guys! @peteonrails, I will be excited to see something like this get into Voxtype. @spencerbull @jacob-vincent-mink happy to help in anyway I can. Please let me know if you need anything further.

tslove923:
Pushed a few more commits since this opened — added file-output streaming support (`voxtype record start --file=path` now works in streaming mode, not just the batch path), a silence-timeout auto-stop fix, and a bug fix for an edge case where a short utterance could write an empty transcript file. Streaming + the OpenVINO backend itself is unchanged from what you've been looking at.

tslove923:
Update: reworked this to build on top of @jacob-vincent-mink's #547 instead of keeping a separate OpenVINO transcriber implementation here.

Comparing the two properly (not just assuming this PR was further along) turned up real reasons #547's `transcribe/openvino_whisper.rs` is the stronger base — SDK auto-discovery, `prepare()`/`on_demand_loading`, far more robust model path resolution, `translate` support, and a more robust `is_multilingual` check than what I had here. So instead of porting those improvements into this PR's module, I rebuilt this PR's actual contribution — the sliding-window streaming engine, CACHE_DIR compiled-model caching, file-output streaming plumbing, and the silence-timeout auto-stop — on top of #547's transcriber.

That work is now a stacked PR targeting #547's branch directly, since it depends on #547 landing first: jacob-vincent-mink#1

This PR (#592) still stands as-is for now, but its OpenVINO transcriber is superseded by #547 + the stacked PR above. Happy to close this one out in favor of that combination whenever works for you both — just didn't want to unilaterally close it.

PR #620, opened 2026-08-28
fix: add remote_send_auto_language for whisper.cpp auto-detect

### LINKED ISSUE #619: [Bug] Remote backend: language = "auto" is silently ignored by whisper.cpp server
## Description

When `[whisper] language = "auto"` is configured with `mode = "remote"`, the request sent to the server omits the `language` multipart field entirely (`src/transcribe/remote.rs`, the `if !self.language.is_auto()` guard in `build_multipart_body`).

That is the correct behavior for the OpenAI API, which treats a missing `language` as auto-detect and rejects non-ISO-639-1 values. But whisper.cpp's example server treats a missing `language` differently: it falls back to whatever `-l` the server was started with, which defaults to `en` (whisper.cpp `examples/server/server.cpp`, default `language = "en"`). The server does accept an explicit `language=auto` and runs real auto-detection in that case.

The result is that voxtype's `language = "auto"` is silently downgraded to English (or whatever the server default is) when talking to a whisper.cpp server. No error is reported anywhere; transcription of non-English audio just comes out wrong.

## Steps to Reproduce

1. Start a stock whisper.cpp server without `-l`, exposing the OpenAI-style path so voxtype can reach it: `whisper-server -m ggml-large-v3.bin --inference-path /v1/audio/transcriptions`
2. Configure voxtype with `mode = "remote"`, `remote_endpoint` pointing at that server, and `language = "auto"`
3. Dictate non-English speech (e.g. French)

## Expected Behavior

The server auto-detects the spoken language, as `language = "auto"` promises.

## Actual Behavior

The server transcribes with `lang = en`. Server logs against whisper.cpp 1.9.1:

- Request without a `language` field: server logs `lang = en`
- Same audio with an explicit `language=auto` field: server logs `lang = auto` followed by `auto-detected language: fr (p = 0.954)`

## Environment

- **OS/Distro**: macOS (daemon built from source, `dev` branch); server: whisper.cpp 1.9.1 on Linux
- **Compositor**: n/a (remote backend bug, platform-independent)
- **Voxtype version**: 0.7.5 (dev)
- **Installation method**: built from source
- **Whisper model**: remote (server-side large-v3)

## Additional Context

Starting the server with `-l auto` is a workaround, but it only helps users who control the server, and it doesn't fix the surprise: voxtype's config says `auto` and the setting silently doesn't apply.

Since there is no single value that means auto-detect on both API families (OpenAI needs the field omitted, whisper.cpp needs an explicit `"auto"`), I see two fix shapes:

1. Always send `language=auto` when configured auto. Simplest, but breaks existing OpenAI + `language = "auto"` setups, since OpenAI rejects non-ISO-639-1 values ("Invalid language ... Language parameter must be specified in ISO-639-1 format").
2. An opt-in config option (e.g. `remote_send_auto_language = true`) that sends the explicit `auto` field, with the default preserving today's omit behavior so no existing installation changes.

I have a PR ready for option 2 and will link it here.

## Other findings in the same area

While digging into this I hit two more problems around the remote backend and whisper config. Mentioning them here for visibility; happy to file them as separate issues and work on fixes if you're open to it:

1. **Remote endpoint path is hardcoded.** `src/transcribe/remote.rs` always posts to `/v1/audio/transcriptions` (or `/v1/audio/translations`), but a stock whisper.cpp server only registers `/inference` (plus `/load` and `/health`), so the documented "Sends audio to a remote whisper.cpp server" doesn't work against a default-configured whisper.cpp without `--inference-path`. A configurable path (or an `/inference` fallback) would fix it.

2. **The macOS SwiftUI settings app still reads/writes the deprecated `backend` key.** `macos/VoxtypeSetup/Sources/Settings/WhisperSettingsView.swift` reads `whisper.backend` and writes `backend =` into `[whisper]`, with no migration in ConfigManager. With a config that uses the preferred `mode = "remote"`, the Backend picker silently shows `local`, and touching it appends a second `backend` line, after which the daemon fails to start with `TOML parse error: duplicate key backend in table whisper`. Looks like the same Swift-frontend/Rust-backend config drift family as #576, but this one corrupts the file instead of failing to write.

musingfox:
## Description

When `language = "auto"`, the remote backend omits the `language` multipart field. OpenAI-compatible endpoints treat that as auto-detect, but whisper.cpp's server falls back to its own `-l` setting (default `en`), silently ignoring voxtype's auto-detect configuration.

This adds an opt-in `[whisper] remote_send_auto_language` option (default `false`) that sends an explicit `language=auto` field instead. The default preserves current behavior, so existing OpenAI setups (which reject non-ISO-639-1 values like `"auto"`) are unaffected.

Verified against whisper.cpp 1.9.1: a field-less request logs `lang = en`, while an explicit `language=auto` logs `lang = auto` followed by `auto-detected language: fr (p = 0.954)`.

## Related Issue

Fixes #619

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## Testing

- [x] I have run `cargo clippy -- -D warnings` with no warnings
- [x] I have run `cargo fmt`

New tests: `test_multipart_body_omits_language_when_auto` (default keeps omitting the field) and `test_multipart_body_sends_auto_language_when_enabled` (opt-in sends `auto`).

Note: `cargo clippy --all-targets -- -D warnings` reports 12 pre-existing errors on macOS targets (daemon.rs, menubar.rs, hotkey_macos.rs, etc.) that are present on a clean `dev` checkout and untouched by this PR; the Linux CI lanes are unaffected.

## Documentation

- [x] I have updated documentation as needed (`docs/CONFIGURATION.md`, commented option in the default config template)
- [ ] No documentation changes are needed

## Pre-merge checklist

- [ ] Wait for `ci-success` to pass before marking Ready for Review (draft mode is encouraged while CI is red)

## Additional Notes

Since there is no single value that means auto-detect on both API families (OpenAI needs the field omitted, whisper.cpp needs an explicit `"auto"`), the option defaults to the current omit behavior and is documented as a whisper.cpp-server switch. If you'd rather flip the default or take a different shape entirely, happy to adjust.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1070,6 +1070,28 @@ remote_endpoint = "http://192.168.1.100:8080"
 remote_timeout_secs = 60  # 60 second timeout for long recordings
 ```
 
+### remote_send_auto_language
+
+**Type:** Boolean
+**Default:** `false`
+**Required:** No
+
+Controls what happens to the `language` field of the request when `language = "auto"`.
+
+By default the field is omitted, which OpenAI-compatible endpoints treat as auto-detect (OpenAI rejects non-ISO-639-1 values such as `"auto"`). whisper.cpp's server instead falls back to its own `-l` setting (default `"en"`) when the field is missing, so auto-detection configured in voxtype is silently ignored. Enable this option to send `language=auto` explicitly.
+
+- For **whisper.cpp server**: Set to `true` (it accepts `"auto"`)
+- For **OpenAI API**: Leave `false` (omitting the field means auto-detect)
+
+**Example:**
+```toml
+[whisper]
+mode = "remote"
+remote_endpoint = "http://192.168.1.100:8080"
+language = "auto"
+remote_send_auto_language = true
+```
+
 ### whisper_cli_path
 
 **Type:** String
--- a/src/config/default_config.rs
+++ b/src/config/default_config.rs
@@ -165,6 +165,12 @@ translate = false
 #
 # Timeout for remote requests in seconds (default: 30)
 # remote_timeout_secs = 30
+#
+# Send an explicit "language=auto" field when language = "auto" (default: false)
+# By default the language field is omitted in auto mode, which OpenAI-compatible
+# endpoints treat as auto-detect. whisper.cpp's server instead falls back to its
+# own -l setting (default "en"); enable this so "auto" is sent explicitly.
+# remote_send_auto_language = true
 
 [output]
 # Primary output mode: "type" or "clipboard"
--- a/src/config/whisper.rs
+++ b/src/config/whisper.rs
@@ -144,6 +144,15 @@ pub struct WhisperConfig {
     #[serde(default)]
     pub remote_timeout_secs: Option<u64>,
 
+    /// Send an explicit `language=auto` field when language = "auto" (default: false)
+    /// OpenAI-compatible endpoints treat a missing language field as auto-detect
+    /// (and reject non-ISO-639-1 values like "auto"), but whisper.cpp's server
+    /// falls back to its own `-l` setting (default "en") when the field is
+    /// missing, silently ignoring voxtype's auto-detect configuration.
+    /// Enable this when using a whisper.cpp server so "auto" is sent explicitly.
+    #[serde(default)]
+    pub remote_send_auto_language: bool,
+
     // --- CLI backend settings ---
     /// Path to whisper-cli binary (optional, searches PATH if not set)
     /// Used when mode = "cli"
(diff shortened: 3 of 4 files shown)


peteonrails:
This landed in the 1.0.1 release branch as `2228250`, with your authorship preserved, and closes #619. It ships in v1.0.1.

Apologies for the roundabout route: the 1.0.1 work was assembled on a release branch rather than merged PR by PR, so your commit was carried across directly instead of this PR being merged. The code is yours and the credit is recorded in the release notes.

Thanks for the fix. `language = "auto"` being silently dropped on the remote backend is the kind of bug that wastes a lot of someone's time before they think to read the source.

peteonrails:
Closing as merged via `2228250` on the 1.0.1 release branch.

PR #621, opened 2026-08-28
fix: Eager processing transcribes chunks with whisper when a non-whisper engi...

### LINKED ISSUE #618: Eager processing transcribes chunks with whisper when a non-whisper engine uses on-demand loading
## Summary

With `engine = "parakeet"`, `[parakeet] on_demand_loading = true`, and `eager_processing = true`, in-flight eager chunks are transcribed by the whisper model on CPU while only the tail goes through Parakeet. The combined output stitches text from two different engines, and the whisper CPU pass dominates latency.

## Observed (rc4, 2026-08-27)

Daemon logs from a 6.2s dictation, engine=parakeet (MIGraphX):

```
INFO Loading Parakeet Tdt model from ".../parakeet-tdt-0.6b-v2"
INFO Loading model 'small' into cache          <- whisper, via ModelManager
INFO Parakeet Tdt model loaded in 1.11s
INFO Finishing eager recording: 6.2s of audio, 0 chunks already transcribed
INFO Transcription completed in 4.30s: "..."   <- WhisperTranscriber, use gpu = 0
INFO Parakeet Tdt transcription completed in 0.13s: "..."
INFO Combined eager transcription: "..."
```

The 4.30s line is `src/transcribe/whisper.rs` — the first chunk ran through whisper-small on CPU.

## Cause

In the eager chunk dispatch (src/daemon.rs, "Populate eager transcriber cache on first poll"), `eager_transcriber` is taken from `transcriber_preloaded`. With on-demand loading nothing is preloaded, so the code falls through to the ModelManager — which is whisper-only — and uses that transcriber for every chunk, regardless of the configured engine. The comment on the fallback says "Whisper engine" but nothing gates it to the whisper engine.

## Expected

Chunks of a Parakeet (or any non-whisper) session must never be transcribed by whisper. When the engine transcriber is not yet available, chunk dispatch should wait for the engine's load task; the finish path already transcribes all pending audio with the engine transcriber, so deferring chunks costs nothing correctness-wise.

## Impact

- Wrong engine silently used for part of the utterance (accuracy and consistency: different casing/punctuation conventions mid-utterance)
- Latency: whisper-small on CPU took 4.30s for a ~6s chunk on a machine whose Parakeet path does it in ~0.15s
- The config combination is reachable through normal setup: the TUI can produce parakeet + on_demand + eager

Found while diagnosing a slow dictation; full log excerpt available.

yhay81:
Fixes #618.

## Summary

Eager processing transcribes chunks with whisper when a non-whisper engine uses on-demand loading

## Validation

- Mechanical gate: +4/-1, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -3409,7 +3409,10 @@ impl Daemon {
                             _ => None,
                         };
                         eager_transcriber = transcriber_preloaded.clone();
-                        if eager_transcriber.is_none() {
+                        if eager_transcriber.is_none()
+                            && self.config.engine
+                                == crate::config::TranscriptionEngine::Whisper
+                        {
                             // Whisper engine: get from model manager
                             if let Some(ref mut mm) = self.model_manager {
                                 match mm.get_prepared_transcriber(model_override) {


peteonrails:
THANK YOU @yhay81! When I filed #618 two days ago I figured I would get around to it shortly....appreciate you getting there first!

PR #659, opened 2026-08-30
Fade media ducking without delaying the microphone

peteonrails:
Supersedes #585 and #586, both by @lunetics, whose commits and work this builds on directly. Resolves the three issues raised in review on #585 so the fade and the cube-root correction can land together.

## What was wrong

**The fade delayed every recording.** `duck_playing_audio` ramped inline, and it's awaited in `suppress_recording_media()` before `start_recording_capture()` — so the microphone waited out the full fade plus seven sequential `pactl` spawns per stream. Push-to-talk clipped first syllables; @peteonrails reproduced it.

**The restore race was a ratchet.** Because restore now runs for the whole fade, a second recording started inside that window enumerated *mid-fade* volumes and stored them as the new originals. Every fast toggle cycle stored a quieter baseline, so media drifted permanently down and never recovered.

**The cube-root fix would have made ducking inaudible on defaults.** #586's analysis is correct — pactl percentages relate to amplitude as (percent/100)³, so the old code cubed the reduction. But the shipped default of 70 was being applied to that cubic scale and actually left 0.70³ = 0.343 of the amplitude. Reinterpreting 70 as amplitude directly turns a -9.3 dB duck into -3.1 dB for everyone on defaults.

## What changed

- The ramp runs on a spawned task, with originals captured before it starts, so capture begins immediately.
- The daemon holds that task so duck and restore serialize. Duck **awaits** a restore in flight, because only its final write puts the true originals back. Restore **aborts** a duck in flight, because restore ends by writing those originals anyway.
- Default `duck_media_volume_percent` moves 70 → 34, keeping the audible depth identical under the corrected meaning. A test pins the relationship so it can't be quietly rounded back.
- Adds `--duck-media-fade-ms`, `VOXTYPE_DUCK_MEDIA_FADE_MS`, and the config-set schema entry. Before this, `voxtype config set audio.duck_media_fade_ms 200` returned "unknown config key" while the neighbouring volume key worked.
- Unshuffles the rustdoc the fade commit displaced onto `FADE_STEP_MS`.

## Upgrade note

Anyone who set `duck_media_volume_percent` **explicitly** needs to cube their old number — an old `50` is a new `13`. Users on the default are unaffected. `CONFIGURATION.md` documents this under a "Changed in 1.1.0" note.

## Testing

`cargo test` 957 passing, `cargo clippy --all-targets --no-deps -- -D warnings` clean, `cargo fmt --check` clean. Verified `voxtype config get audio.duck_media_fade_ms` now returns a value and `--duck-media-fade-ms` appears in help.

Not manually smoke-tested against live audio — worth one pass with `duck_media = true` and music playing before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -714,7 +714,7 @@ We want to hear from you! Voxtype is a young project and your feedback helps mak
 - [KaiStarkk](https://github.com/KaiStarkk) - Post-process trim and fallback_on_empty options
 - [graysky](https://github.com/graysky2) - Flash attention config fix
 - [OldJobobo](https://github.com/OldJobobo) - Quickshell OSD theming, Omarchy theme state path support, configurable media ducking, degenerate Whisper transcript retry; co-owns `quickshell/`
-- [Matthias Breddin](https://github.com/lunetics) - IBus non-ASCII reordering troubleshooting, Rust 1.98 clippy fix, media ducking review
+- [Matthias Breddin](https://github.com/lunetics) - IBus non-ASCII reordering troubleshooting, Rust 1.98 clippy fix, media ducking fade, cubic-scale ducking volume correction
 
 ## License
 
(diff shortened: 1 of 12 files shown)


lunetics:
Lgtm, the old branches can then be closed right? Happy to see that go live.

PR #686, opened 2026-09-01
fix: KDE Wayland output driver via kwtype

### LINKED ISSUE #649: [Feature] KDE Wayland output driver via kwtype
## Summary

We have no KDE-specific output driver. On KDE Wayland our chain falls through to whatever else probes successfully, and wtype in particular cannot work there — KWin does not implement `zwp_virtual_keyboard_manager_v1`.

[Handy](https://github.com/cjpais/Handy) ships `kwtype` support (`src-tauri/src/clipboard.rs:202-206`), which uses KDE's Fake Input protocol and handles umlauts correctly.

## Scope

A `kwtype` driver in `src/output/` alongside the existing wtype/dotool/ydotool/clipboard modules, probed ahead of the others when the session is KDE Wayland. Same trait, same fallback semantics.

## Related

#650 covers making the wtype probe compositor-aware, which is the other half of getting KDE right. That one is smaller and worth doing first.

## Why 1.2.0

KDE Plasma is a real share of Linux desktops and this closes a genuine hole, but it is additive rather than a defect — nothing is broken that was previously working, and the clipboard fallback still delivers text today.

yhay81:
Fixes #649.

## Summary

[Feature] KDE Wayland output driver via kwtype

## Validation

- Mechanical gate: +126/-9, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.

### DIFF
--- a/src/config/output.rs
+++ b/src/config/output.rs
@@ -350,6 +350,8 @@ pub enum OutputMode {
 #[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 pub enum OutputDriver {
+    /// kwtype - KDE Plasma Wayland via KWin's Fake Input protocol
+    Kwtype,
     /// wtype - Wayland-native via virtual-keyboard protocol, best Unicode/CJK support
     Wtype,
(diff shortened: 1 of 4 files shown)


Kiraprint:
Btw, I tested it and it's working on KDE.
All file changes are clear.
Thx to the author!

PR #688, opened 2026-09-01
fix: wire Cohere models into setup --download

### LINKED ISSUE #687: voxtype setup --download does not know about Cohere models, blocking the Cohere engine
## What's wrong

The daemon supports the `cohere` engine (listed in `voxtype info engines`, and it will try to load a Cohere model if configured), but there is no working way to actually download that model through the CLI.

## Steps to reproduce

1. Switch to the ONNX binary variant (`voxtype setup onnx --enable`, as root) — required, since the default AVX2-only binary fails with "Cohere engine requested but voxtype was not compiled with --features cohere".
2. Set `engine = "cohere"` in `config.toml`, plus a `[cohere]` section with `model = "cohere-transcribe-q4f16"` (needed too — without it, the daemon fails with "Cohere engine selected but [cohere] config section is missing").
3. Restart the daemon. It fails:
   ```
   Error: Transcription error: Model not found: Cohere encoder not found: /home/<user>/.local/share/voxtype/models/cohere-transcribe-q4f16/encoder_model.onnx
     Download from https://huggingface.co/onnx-community/cohere-transcribe-03-2026-ONNX
     Run 'voxtype setup' to download models.
   ```
4. Following that instruction: `voxtype setup --download --model cohere-transcribe-q4f16` fails too:
   ```
   Error: Unknown model 'cohere-transcribe-q4f16'.
     Whisper: tiny, tiny.en, base, base.en, small, small.en, medium, medium.en, large-v3, large-v3-turbo
     Parakeet: parakeet-tdt-0.6b-v2, parakeet-tdt-0.6b-v2-int8, parakeet-tdt-0.6b-v3, parakeet-tdt-0.6b-v3-int8, parakeet-unified-en-0.6b
     SenseVoice: sensevoice-small, sensevoice-small-fp32
   ```

So the model catalog behind `--download` has no entry for Cohere, and there's no other documented/scripted way to get the files into place. Right now Cohere is completely unreachable outside of `voxtype configure`'s interactive picker (untested, may or may not handle it — haven't found a non-interactive path).

## Expected behavior

`voxtype setup --download --model cohere-transcribe-q4f16` (and the other Cohere variants: q4, int8, fp16) should fetch the right files from `onnx-community/cohere-transcribe-03-2026-ONNX` on Hugging Face into `~/.local/share/voxtype/models/cohere-transcribe-q4f16/`, matching what the loader expects (`encoder_model.onnx` etc.).

## Environment

- voxtype-bin 1.0.0-2 (Omarchy repo build)
- ONNX (AVX2) binary variant, x86_64
- Arch/Omarchy, kernel 7.1.9-arch1-2

felipecpaiva:
## What

`voxtype setup --download --model <cohere-variant>` rejected every Cohere model name with "Unknown model", even though the daemon fully supports the `cohere` engine and the download/validate/config-write plumbing for it already existed in `model.rs` (used by the interactive `voxtype setup model` picker). This left Cohere completely unreachable through the non-interactive CLI, which is what the docs and the daemon's own error message point users to.

## Fix

Added a `Cohere` branch to `ModelKind` / `classify_model_override` and to `run_setup`'s dispatch in `src/setup/mod.rs`, mirroring the existing Parakeet branch exactly (feature-gated on `cohere`, same download/activate/not-yet-downloaded flow).

Added four small helpers in `src/setup/model.rs`, mirroring the SenseVoice ones:
- `find_cohere_model` / `is_cohere_model` / `cohere_dir_name` — dual match on short name (`q4f16`) or directory name (`cohere-transcribe-q4f16`), same pattern as `find_sensevoice_model`, since `voxtype info models` and the daemon's "download from" error print the directory form.
- `cohere_setup_model_names` — feeds the "Unknown model" error message's model list, same as the other engines.
- `set_cohere_config` — quiet variant of `update_config_cohere` for the `--activate` flag, mirroring `set_parakeet_config`.

`download_cohere_model` now resolves through `find_cohere_model` instead of only matching the short name, so it accepts either form too.

## Testing

- `cargo test --features cohere` — all existing tests pass (no new tests added; this is a routing fix with no new logic worth unit-testing beyond what the existing Parakeet/SenseVoice tests already cover the same way).
- `cargo fmt -- --check` — clean.
- `cargo clippy --all-targets --no-deps -- -D warnings` (default features) — clean.
- `cargo clippy --features cohere --all-targets --no-deps -- -D warnings` — fails, but identically on a clean checkout of `dev` before this change (pre-existing warnings in `src/transcribe/cohere.rs` and `src/transcribe/ctc.rs`, unrelated to this diff — confirmed via `git stash` on this branch). Not touched here per the "stay inside the files your change already touches" guidance.
- Could not exercise the actual multi-GB download end-to-end (no network egress for a file that size in this environment), but traced the fix through the same code path the interactive picker already uses successfully (`download_artifact` + `validate_cohere_model`), so behavior should match `voxtype setup model`'s Cohere option exactly — it now just also works with `--download --model <name>`.

Fixes #687


felipecpaiva:
Rebased onto current `dev` — force-pushed, single commit, and GitHub now reports the PR `MERGEABLE` again.

`dev` had moved 46 commits and the conflicts were all one shape: the OpenVINO engine landed in exactly the places this PR adds a Cohere arm. Five hunks in `src/setup/mod.rs`, every one resolving to "keep both" — the `ModelKind` variant, the `classify_model_override` chain, the two halves of the `bail!` message, and the `run_setup` dispatch block. `src/setup/model.rs` merged clean and its 47-line change is byte-identical to before.

Cohere is chained last in `classify_model_override`, after OpenVINO. Checked for shadowing: Cohere's short names are `q4f16`/`q4`/`int8`/`fp16`, every OpenVINO name carries a Whisper size prefix (`small-int8`, `base-fp16`), Whisper's are bare sizes, Parakeet's are `parakeet-*`, SenseVoice's are `small`/`small-fp32`. No collisions, and being last means Cohere can only lose one, never shadow another engine — consistent with the "Whisper wins name collisions" note you added.

The fix is still needed on current `dev`: `git grep -i cohere dev -- src/setup/mod.rs` returns nothing, so `setup --download --model <cohere>` still fails with "Unknown model" even though `model.rs` already has the whole download/validate/config-write path that the interactive `setup model` picker uses.

Verified locally against `.github/workflows/ci.yml`: `cargo fmt --check` clean, `cargo clippy --all-targets --no-deps -- -D warnings` clean, `cargo test` 1115 passed / 0 failed / 2 ignored. Also ran `cargo check --features cohere`, because CI never compiles that feature — the entire fix body is `#[cfg]`-ed out of every CI job, so nothing in the matrix actually type-checks it.

Two notes, one for you and one for me:

- Both workflow runs on this push came back `action_required` (first-time-contributor gate), so there is no green check on the PR until someone approves the run.
- I did not exercise the Cohere download itself — it is a multi-GB fetch — so "the routing now reaches the downloader" is verified by compilation, not by a run. Same as the original PR.

One deliberate non-change: `left_damaged` is set only in the Whisper branch, and OpenVINO, SenseVoice and Parakeet all ignore it. I left Cohere consistent with those three rather than extending it here. Happy to do that as a separate PR covering all four if you want it.

PR #698, opened 2026-09-02
Wait for OpenVINO NPU compilation during model setup

jacob-vincent-mink:
## Summary

- eagerly compile downloaded OpenVINO models when the configured device is NPU
- block setup until strict NPU pipeline initialization succeeds and a cache blob exists
- avoid GPU/CPU fallback masking an NPU compilation failure
- retry NPU preparation when model files remain after an earlier failed compile
- show users that first-time compilation may take several minutes

## Verification

- `cargo check`
- `cargo check --features openvino-whisper`
- focused NPU precompile decision and cache-blob detection tests

### DIFF
--- a/src/setup/mod.rs
+++ b/src/setup/mod.rs
@@ -640,6 +640,14 @@ pub async fn run_setup(
                 model_path.exists() && model::validate_openvino_model(&model_path).is_ok();
 
             if model_valid {
+                // A failed first compile leaves valid downloaded IR files in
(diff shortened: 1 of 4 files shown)


jacob-vincent-mink:
tested local with the big models, works good. @tslove923 you interested in running a quick pass? @spencerbull

jacob-vincent-mink:
worked for @spencerbull too via i ran it on his computer from source

Issue #244, opened 2026-03-01
Feature: single daemon for hotkey dictation + OpenAI-compatible local STT API


krystophny:
## Problem

Running desktop hotkey dictation and API STT as separate daemons increases operational complexity and failure modes.

## Proposal

Add a service mode to `voxtype daemon` that starts a local OpenAI-compatible transcription API in parallel with the existing daemon loop.

### Goals

- Single daemon process for:
  - existing hotkey-driven dictation
  - local API transcription
- OpenAI-compatible endpoint:
  - `POST /v1/audio/transcriptions`
- Health endpoint:
  - `GET /healthz`
- Loopback-only by default (`127.0.0.1`)
- No built-in auth in v1 (assume trusted localhost)

### Language behavior

- General language support remains configurable.
- Default constrained auto set for local deployment/tests: `de,en`.

### Non-goals (v1)

- Public network exposure
- Built-in auth/mTLS
- TTS or chat endpoints

## Why

This enables a clean architecture where upstream voxtype is the single local speech daemon, while another service (for example Tabura) handles external auth/proxy and forwards requests to localhost.

## Acceptance criteria

- `voxtype daemon` can run hotkey flow and API flow concurrently.
- API transcriptions work while daemon is active.
- Existing daemon behavior remains unchanged when service mode is disabled.
- Integration tests cover multipart transcription endpoint and concurrent operation.

PR #706, opened 2026-09-02
Voxtype Cloud pilot: OpenAI-compatible STT Worker + remote meeting diarization

peteonrails:
Pilot for a hosted Voxtype Cloud offering and the protocol seed for #244.

## What's in here

**`cloud/` — a Cloudflare Worker (new wrangler project)** implementing an OpenAI-compatible `POST /v1/audio/transcriptions` (`json` | `verbose_json`) backed by Deepgram Nova-3 on Workers AI, plus `GET /v1/models`. One documented voxtype extension: a `diarize=true` form field returns per-speaker-turn segments (`speaker`/`confidence` on segments and words in `verbose_json`). `whisper*` model names route to whisper-large-v3-turbo (1/10 the price, no diarization). Single-key Bearer auth with constant-time compare; the key-lookup seam is ready for per-user keys. `/v1/chat/completions` and `/v1/realtime` are reserved routes. Audio/transcripts are never logged.

**Client — remote meeting diarization.** `[meeting.diarization] backend = "remote"` (config or `voxtype meeting start --diarization remote`) makes meetings request speaker-diarized segments from the remote whisper server: `with_meeting_mode_overrides()` sets an internal `whisper.remote_diarize` flag, `RemoteTranscriber` gains a `transcribe_timed()` override parsing `segments[]` (graceful degrade against plain OpenAI servers), `TimedSegment` gains optional `speaker`/`confidence`, and `ChunkProcessor` maps them to `speaker_id` (mic → "You", loopback → `SPEAKER_NN`, so `meeting label` and the integer-keyed labels table work unchanged). No local diarizer runs when the gate is satisfied. Dictation is untouched — plain remote mode against the Worker already works with zero client changes (verified with the stock 1.0.0 binary).

## Verified

- Worker: 14 vitest tests (turn grouping, fixture parsing, REST-fallback request shape, error hints); curl matrix against the deployed Worker (auth 401, missing file 400, bad format 400, models list, whisper json transcription, nova-3 502-with-hint).
- Client: `cargo test` (1,075 lib tests incl. 14 new), `clippy -D warnings`, `fmt`. Old TOML without the new field parses; the flag defaults off.
- E2E: stock installed voxtype 1.0.0 **and** this branch's binary both transcribe through the deployed Worker (`--engine whisper --whisper-mode remote`), ~2 s round trip on 15 s audio.

## Upstream blockers (documented, not ours)

1. **Workers AI binding rejects nova-3 audio input** (cloudflare/workerd#5082) — every body encoding fails with error 5006. The Worker falls back to the documented REST path (verified live) when a `CLOUDFLARE_API_TOKEN` secret (Workers AI Run scope) is set; the binding takes over automatically when fixed.
2. **Nova-3 diarization on Workers AI tags every word speaker 0**, including on Deepgram's own two-speaker demo files (`diarize=true` demonstrably engages — speaker fields appear only with it). Until Cloudflare fixes the hosted model, cloud diarization returns a single speaker; the whole pipeline is in place and needs no changes when it's fixed.

## Known pilot limitations

Chunk-local speaker numbering (1:1 calls fully stable; raise `chunk_duration_secs` for multi-party), `prompt` ignored by the Worker, batch only, single shared API key, no retry on failed chunks. The privacy shift (audio leaves the machine) is called out in CONFIGURATION.md, USER_MANUAL.md, and TROUBLESHOOTING.md.

### DIFF
--- a/cloud/.gitignore
+++ b/cloud/.gitignore
@@ -1,0 +1,4 @@
+node_modules/
+.wrangler/
+.dev.vars
+worker-configuration.d.ts
--- a/cloud/README.md
+++ b/cloud/README.md
@@ -1,0 +1,119 @@
+# Voxtype Cloud (pilot)
+
+An OpenAI-compatible speech-to-text API on Cloudflare Workers, backed by
+Workers AI. Transcription runs on `@cf/deepgram/nova-3` (with optional speaker
+diarization) or `@cf/openai/whisper-large-v3-turbo`. This Worker is the pilot
+for a hosted Voxtype Cloud offering and the seed of the planned local
+OpenAI-compatible STT API (#244) — voxtype's existing remote Whisper client
+speaks this protocol without modification.
+
+## Endpoints
+
+All endpoints require `Authorization: Bearer <key>`.
+
+### `POST /v1/audio/transcriptions`
+
+OpenAI-compatible multipart form:
+
+| Field | Notes |
+|---|---|
+| `file` | required; the audio (WAV/MP3/OGG — content type taken from the part) |
+| `model` | `nova-3` (default) or any `whisper*` name → whisper-large-v3-turbo |
+| `language` | Nova-3 set only (en/es/fr/de/hi/ru/pt/ja/it/nl + regional variants, or `multi`); anything else is omitted |
+| `response_format` | `json` (default) or `verbose_json` |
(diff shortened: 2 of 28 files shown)


peteonrails:
Experimental, not supposed to be a PR yet

PR #707, opened 2026-09-03
streaming: fix select! starvation stalling stop under slow inference

tslove923:
## Summary

The sliding-window streaming engine's event loop raced `samples_rx.recv()` (chunk/Eof), `cancel_rx`, and `ticker.tick()` in an **unbiased** `tokio::select!`. Once inference falls behind the tick interval, the ticker is essentially always ready by the time control returns to the `select!` — an unbiased `select!` picks pseudo-randomly among ready arms, so a real stop signal (mic stopped, `samples_rx` closing) could keep losing that draw to "fire another tick" indefinitely.

## Found live

On NPU-backed OpenVINO streaming: a session whose buffer grew past ~13s fell behind the 0.8s tick (inference routinely over 1s), and the channel-close signal sat unprocessed for **33+ seconds** after the user pressed stop — the loop kept re-transcribing the same frozen buffer once a second at full NPU load, until the unrelated hard `max_duration_secs` recording cap force-ended it.

## Fix

Mark the `select!` `biased` and order `Chunk`/`Eof` and `Cancelled` ahead of `Tick`, so a real chunk or a closed/cancelled channel always wins over firing another expensive re-transcription whenever both are ready. A queued stop is now caught on the very next loop iteration once the in-flight `transcribe()` call returns, regardless of how far behind inference has fallen or how many ticks are backed up.

Also adds debug/trace logging around this loop (chunk drains, Eof/Cancelled/on_tick-error transitions, and a warning when 2+ consecutive ticks fire without draining a chunk) so a recurrence — or any future regression in this area — is visible in logs instead of only inferable from buffer-growth numbers in trace output.

## Verification

Verified live end-to-end on real NPU hardware: a 22.8s dictation with inference regularly taking 3+ seconds per tick now ends **~4.3s** after the stop key is pressed (almost entirely one unavoidable `final_flush` call), down from **33+ seconds** before the fix, with zero "consecutive ticks without draining a chunk" warnings during the drain.

Marked as draft while I finish writing an automated regression test for this (an earlier attempt turned out to be timing-flaky and was dropped rather than shipped unverified).

### DIFF
--- a/src/transcribe/sliding_window.rs
+++ b/src/transcribe/sliding_window.rs
@@ -279,26 +279,65 @@ impl StreamingTranscriber for SlidingWindowStreamingTranscriber {
                 Cancelled,
             }
 
+            // How many *consecutive* ticks fired without a single chunk
+            // being drained in between. Purely diagnostic (see the WARN
+            // below) — doesn't affect the fix itself, just makes the
+            // starvation pattern visible in logs if it ever recurs.
+            let mut consecutive_ticks_without_drain: u32 = 0;
+
             loop {
                 let outcome = runtime.block_on(async {
+                    // `biased` makes `select!` check arms top-to-bottom
+                    // instead of picking a ready one at pseudo-random.
+                    // Chunk/Eof/Cancelled must win over Tick whenever both
+                    // are ready — otherwise, once inference falls behind
+                    // interval_s, the ticker arm is *permanently* ready
+                    // (MissedTickBehavior::Skip just collapses the backlog
+                    // to one ready tick, it doesn't stop being ready) and
+                    // can keep winning the random tie-break against a
+                    // stop/EOF signal indefinitely. Found live: a session
+                    // whose buffer grew past ~13s fell behind the 0.8s
+                    // tick, and the mic-stop signal (samples_rx closing)
+                    // sat unprocessed for 33+ seconds — the loop kept
+                    // re-transcribing the same frozen buffer once a second
+                    // at full NPU load, spinning until the unrelated hard
+                    // `max_duration_secs` recording cap force-ended it.
+                    // Biasing toward Chunk/Eof/Cancelled means a queued
+                    // stop is picked up on the very next loop iteration
+                    // once the in-flight transcribe() call returns,
+                    // regardless of how many ticks are backed up.
                     tokio::select! {
+                        biased;
+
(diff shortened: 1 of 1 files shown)


peteonrails:
@tslove923 This looks like it's going to be important: please ping me if you need any help! Side note: Do you have a GPG key you could use to set up commit-signing? I'm trying to increase the supply chain trust and will be turning on required signing on the repo soon. 

image

src/transcribe/sliding_window.rs
```diff
@@ -279,26 +279,65 @@ impl StreamingTranscriber for SlidingWindowStreamingTranscriber {
...
+                    // regardless of how many ticks are backed up.
                     tokio::select! {
+                        biased;
+
                         chunk = samples_rx.recv() => match chunk {
                             Some(c) => Outcome::Chunk(c),
                             None => Outcome::Eof,
                         },
-                        _ = ticker.tick() => Outcome::Tick,
                         _ = &mut cancel_rx => Outcome::Cancelled,
```
jacob-vincent-mink, line 316: With biased now, do we want to cancel before drain? Currently, would make cancelled come "late", right?

src/transcribe/sliding_window.rs
```diff
@@ -279,26 +279,65 @@ impl StreamingTranscriber for SlidingWindowStreamingTranscriber {
...
+                    // sat unprocessed for 33+ seconds — the loop kept
+                    // re-transcribing the same frozen buffer once a second
+                    // at full NPU load, spinning until the unrelated hard
+                    // `max_duration_secs` recording cap force-ended it.
+                    // Biasing toward Chunk/Eof/Cancelled means a queued
+                    // stop is picked up on the very next loop iteration
+                    // once the in-flight transcribe() call returns,
+                    // regardless of how many ticks are backed up.
                     tokio::select! {
+                        biased;
```
jacob-vincent-mink, line 310: Thoughts on a regression test: fill the sample queue while running fake inference, close it, then verify no further periodic tick occurs before EOF/final flush.

tslove923:
Superseding this with #716 — the select! fix grew into a larger rearchitecture of the same engine while chasing related stalling/correction bugs, so this fix is folded back into that branch (last commit, incorporating @jacob-vincent-mink's cancel-before-drain suggestion and the regression test proposed above) rather than force-pushed over this narrower PR. Closing in favor of #716.

PR #716, opened 2026-09-03
Streaming: unified alignment rearchitecture + select! starvation fix

tslove923:
I found some issues with my previous streaming feature when testing the larger models. I had deepseek-v4-flash take a stab at creating a new architecture and it seems to have done a pretty good job. Opting to submit this instead of #707. 

Supersedes and closes #707 — that PR was scoped narrowly to the `select!`
starvation fix, but this branch grew into a larger rearchitecture of the same
engine while investigating related stalling/correction bugs, and this fix is
now folded back in on top of it (see the last commit). Opening fresh rather
than force-pushing over #707 so @jacob-vincent-mink's existing review comments
there stay attached to the diff they were made against.

## What's in this PR

**Unified diff/alignment engine** (`d874d5f`). Replaces the sliding-window
engine's fragmented diff machinery — separate growing-mode prefix match,
sliding-mode tail-delta, and revision-mode paths each with their own state —
with one word-based alignment that powers both commit policies:

- **Stalls fixed**: the old whole-tail stability gate is gone. Type-then-correct
  (now the default) types the tail immediately and corrects per-word via
  `Replace`; a single flickering word is revised every tick instead of
  blocking all text after it. The conservative two-pass gate remains as an
  opt-in (`revision_mode = false`) and is now per-word too.
- **Punctuation-sensitive**: a word that changes only its trailing
  punctuation (`"it."` → `"it?"`) is corrected with a 1-char backspace
  instead of being dropped as "no change" or retyping the whole tail.
- **Premature question marks fixed**: when a new `?` lands on the tail's
  last word, a stale `?` within `question_mark_lookback_words` behind it is
  removed, so `"what time is it? tomorrow afternoon?"` becomes `"what time
  is it tomorrow afternoon?"` — one question mark, on the true end.
- **Per-word, stability-gated promotion**: a word leaves the revisable tail
  only when stable for `stability_passes` ticks *and* beyond
  `revision_lag_words` from the tail's end, so a flickering word never gets
  stuck confirmed on the wrong text.

New config knobs (defaults preserve old behavior except the intentional
`revision_mode` flip): `question_mark_lookback_words` (3), `stability_passes`
(2), `revision_lag_words` (4). `revision_mode` now defaults to `true` — the
shared `[streaming]` section and the deprecated per-engine `streaming_*`
fields both honor it.

**Anchor fix** (`7b43616`). The alignment only matched a suffix of the
committed tail against a *prefix* of the fresh transcription — correct in
sliding mode and short growing sessions, but once committed text exceeds
`SLIDING_DIFF_LOOKBACK_WORDS` the committed tail is no longer a prefix of
curr (it starts mid-sentence), so a single coincidental word anchor could
re-type an entire already-committed sentence on top of itself. Found live as
literal duplication: every tick after ~20 committed words, a REVISE
backspaced a few chars and retyped the whole sentence. Now every start
position in curr is tried against every suffix start of the committed tail,
and the anchor consuming the most committed words wins.

**TUI: Streaming section** (`f51d09b`, `678e647`). New `configure` section
exposing the shared `[streaming]` tuning knobs, plus a per-engine on/off
toggle that reads/writes `[<engine>] streaming` for whichever engine is
currently active (whisper, openvino, parakeet, soniox — engines without a
`streaming` field show a read-only toggle explaining why). Sits at the
bottom of the sidebar. Defaults match the engine's new `revision_mode = true`
default, and the guidance text presents the conservative gate as the opt-in
rather than the other way around.

**select! starvation fix, restored + improved** (`57684f9`). This branch
started from `dev` before #707 merged, and the rearchitecture rewrote the
event loop without carrying that fix forward — so the original bug (an
unbiased select! lets an always-ready `ticker.tick()` win against a closing
sample channel once inference falls behind `interval_s`, stalling a real
stop signal for 30+s) was back. Re-applied with `biased;`, but with a
different arm order than #707's original fix per @jacob-vincent-mink's
review comment there: `Cancelled` now comes *first*, ahead of `Chunk/Eof`,
not after — `cancel_streaming_to_idle`'s own contract is "abort promptly, no
flush required," so an explicit cancel (user cancel action, or SIGTERM
shutdown) should win even over a chunk that's already queued and ready to
drain, rather than sitting behind whatever backlog is in the channel. `Eof`
still can't itself starve (mpsc only yields it once every queued chunk is
drained), so `Chunk/Eof` stays one arm, ahead of `Tick`.

Also adds the regression test Jacob proposed on #707: fill the sample queue
while a slow fake-inference call is in flight, close the channel, and assert
no extra periodic tick fires between the queue draining and EOF/final_flush.
Verified this actually catches the regression (not tautological): temporarily
reverting to the old `Tick`-first order made it fail with 6 `transcribe()`
calls instead of the expected 2.

## Testing

- `cargo test`: 1073/1073 passing, 0 clippy warnings, `cargo fmt --check`
  clean on all lines this PR touches.
- Live-tested on real NPU dictation (OpenVINO GenAI, `medium-int4`/NPU):
  long continuous recordings commit smoothly with no stalls, punctuation
  corrections are clean single-character backspaces, stop/cancel is
  responsive. No warnings or errors in trace-level logs across multiple
  sessions.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1241,9 +1241,12 @@ This section is entirely optional. Omitting it (or omitting individual fields in
 | `max_buffer_secs` | Float | `29.0` | Max buffered audio (seconds) before the window slides (drops old samples to respect the model's context limit) |
 | `min_speech_rms` | Float | `0.005` | Skip transcription while whole-buffer RMS is below this |
 | `min_audio_secs` | Float | `1.0` | Min buffered audio (seconds) before the first partial is attempted |
-| `partial_min_words` | Integer | `1` | Min new stable words before a delta is committed/typed |
+| `partial_min_words` | Integer | `1` | Min new stable words before a delta is committed/typed (conservative gate only — type-then-correct ignores it and types everything immediately) |
 | `type_partials` | Boolean | `true` | Type committed deltas live at the cursor, vs. commit whole segments at once |
-| `revision_mode` | Boolean | `false` | Experimental: type immediately and correct via backspace if wrong, instead of waiting for agreement. See `openvino.streaming_revision_mode` in the `[openvino]` section for the full explanation of the trade-off — it applies identically here |
+| `revision_mode` | Boolean | `true` | **Type-then-correct (default):** type immediately and correct via backspace + retype if a later pass disagrees, instead of withholding the tail until it's stable across `stability_passes` ticks. Setting this to `false` opts back into the legacy conservative gate (never types something wrong, but lags by ~`stability_passes` ticks and can visibly pause) |
+| `question_mark_lookback_words` | Integer | `3` | When a new "?" lands on the tail's last word, scan back this many words for a stale earlier "?" and remove it, so a still-open question keeps exactly one "?" on its true end (e.g. "what time is it? tomorrow afternoon?" becomes "what time is it tomorrow afternoon?"). Capped at `revision_lag_words` so it only ever revises still-revisable text |
+| `stability_passes` | Integer | `2` | Consecutive ticks a word must be present (same text and punctuation) before it's promoted out of the revisable tail (type-then-correct mode) or typed at all (conservative gate) |
+| `revision_lag_words` | Integer | `4` | How many trailing words of the tail stay revisable in type-then-correct mode; anything older is permanently confirmed and can never be revised again, bounding how far a correction can reach |
 
 **Example:**
 ```toml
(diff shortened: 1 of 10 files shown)


tslove923:
## Streaming engine tested across 3 machines, 10 backend/device combos

Ran the sliding-window streaming engine from this PR against a fixed 25s
reference clip, on whisper.cpp and OpenVINO, across every CPU/GPU/NPU
combination three different machines could support. Full setup + reproduction
steps + the exact reference audio are pushed to
[`bench/streaming-silicon`](https://github.com/tslove923/voxtype/tree/bench/streaming-silicon/bench/streaming-silicon)
on this fork — point another agent at that directory and it can rerun this
exact test on a new machine without rediscovering any of the setup gotchas
below.

**TL;DR: yes, CPU-only whisper.cpp streams correctly** (the original
question this testing set out to answer) — incremental partials land live,
revisions resolve cleanly, no stalling. OpenVINO is substantially faster
across every device it runs on, and Lunar Lake's NPU is the fastest and one
of the most power-efficient combos tested, on any machine.

All three machines are recent, fresh Omarchy installs — this doubles as a
read on what setting voxtype up actually looks like on Omarchy's default
state, not just a hardware comparison on a machine that's already been
tuned.

### Machines

| | CPU | Silicon | Notes |
|---|---|---|---|
| archspectre | i7-1165G7 | Tiger Lake | laptop, Iris Xe iGPU, no NPU |
| archmachine | i5-13600K | Raptor Lake | desktop; iGPU is SR-IOV-split and shares silicon with a Windows VM — treat its GPU numbers as informative, not representative |
| archultra | Core Ultra 7 258V | Lunar Lake | laptop, real NPU (`/dev/accel/accel0`) |

### Results (mean of runs 2–3, run 1 excluded — pays one-time cold-start/graph-compile cost)

| Machine | Backend/Device | Avg tick infer | RTF | WER | Package power |
|---|---|---|---|---|---|
| archspectre | whisper/CPU | 2522ms | 0.578 | 0.314 | 13.8W |
| archspectre | openvino/CPU | 553ms | 0.057 | 0.163 | 13.9W |
| archspectre | openvino/GPU | 457ms | 0.050 | 0.163 | 13.7W |
| archmachine | whisper/CPU | 927ms | 0.120 | 0.163 | 66.4W |
| archmachine | openvino/CPU | 217ms | 0.021 | 0.151 | 40.4W |
| archmachine | openvino/GPU* | 429ms | 0.051 | 0.151 | 36.7W |
| archultra | whisper/CPU | 2224ms | 0.330 | 0.186 | 9.9W |
| archultra | openvino/CPU | 670ms | 0.072 | 0.151 | 9.6W |
| archultra | openvino/GPU | 487ms | 0.048 | 0.163 | 8.0W |
| archultra | **openvino/NPU** | **223ms** | **0.023** | 0.163 | 8.8W |

\* SR-IOV caveat above applies.

RTF (real-time factor) must stay under 1.0 to keep up with live speech — every
combo tested does, with wide margin.

### Findings

- **whisper.cpp CPU streaming works correctly on every machine** — the
  original question. Incremental partials land, revisions resolve, no
  stalling observed in any run.
- **OpenVINO is 4–10x faster than whisper.cpp across the board**, and
  consistently *more* accurate too (WER ~0.15–0.16 vs whisper.cpp's
  0.16–0.31) — surprising given `base.en-int8` is quantized against
  whisper.cpp's unquantized `base.en`.
- **NPU (Lunar Lake) wins outright**: fastest inference of any combo on any
  machine (RTF 0.023), with power usage between GPU (lowest) and CPU
  (highest) on the same chip.
- **Revision churn — the metric that actually tests this PR's stated
  premise** (fixing "stalling/correction bugs") — sat around 20-22
  REVISE events per run across nearly every OpenVINO combo on every machine,
  vs 8-10 for whisper.cpp/CPU. Final accuracy was still better on the
  OpenVINO side despite the higher churn count, and GPU vs CPU openvino
  produced near-identical revision counts on the same machine — suggests
  revision timing is driven by the audio's own structure once an engine is
  comfortably keeping up (RTF well under 1.0), not by which engine is running
  it. Worth a closer look if revision *frequency* itself matters for the UX,
  separate from whether the final text is stable.
- **Package power (RAPL) doesn't cleanly separate CPU vs GPU inference on
  integrated silicon** — on archspectre, `pkg_power_w` came back nearly
  identical (13.7–13.9W) whether inference ran on CPU cores or the iGPU,
  because "package" covers both under one power domain. Battery draw (total
  system, laptops only) showed more separation but is noisy — a legitimate
  cross-check, not a precise measurement.

### A few real setup bugs found along the way (now fixed in the toolkit, not code in this PR)

- `pcm-power` (Intel PCM) hardcodes a per-CPU-model whitelist and silently
  produces empty output on models it doesn't recognize (hit on Tiger Lake) —
  switched to reading the kernel's own RAPL powercap sysfs directly, which
  worked on all three machines.
- voxtype's `[audio] device` config doesn't accept a named PipeWire monitor
  source string directly (only `pipewire`/`default`) — routing a fixed
  reference clip requires swapping the *system* default source instead.
- The OpenVINO GenAI C/C++ SDK isn't in pacman, the AUR, or the
  `openvino-genai` pip wheel — has to be version-matched and pulled from
  Intel's storage bucket by hand.
- `[whisper] streaming` defaults to `false` (unlike `[openvino] streaming`,
  usually already on from prior setups) — easy to silently benchmark
  whisper.cpp in single-shot mode by mistake.

Full detail on all of the above, plus the exact commands, is in the toolkit
README.

PR #736, opened 2026-09-07
Quickshell OSD: recording countdown to audio.max_duration_secs

guidomb:
Adds a seconds-left readout plus depleting time bar to the Quickshell OSD so users can see the auto-stop approaching.

- Counts down from `audio.max_duration_secs` (FileView-synced from ~/.config/voxtype/config.toml, default 60), ticks at 4Hz while recording, resets on record/idle transitions
- Big label (33px, vertically centered right) + 2px bottom time bar; both turn `warning` at <=10s and `error` at <=5s via style roles
- Hidden when idle and in tile/orb/custom layouts; bar mic / Omarchy shell untouched

Motivation: with the default 60s safety cap you only discover the cutoff when transcription stops. This makes remaining time visible next to the waveform.

Tested locally on Omarchy Quattro (voxtype 1.0.1, quickshell frontend, palette omarchy): hold F9 past 50s, observe 10s yellow, 5s red, clean daemon restart with no QML errors.

Happy to rework this as `osd.show_remaining` or a `countdown` visual-layer type if preferred.


OldJobobo:
Thanks, this is useful. I tried the preview and like having the cutoff visible. My only concern is sourcing the duration: reading the config directly misses CLI/environment overrides, and the regex can pick up commented values. Could we use the running daemon’s effective recording limit so the countdown matches the actual auto-stop?

guidomb:
@OldJobobo  Good catch - fixed by threading the daemon's effective value through instead of parsing the file:

- `RuntimeOsdStyle` gains `max_duration_secs`; the supervisor forwards it as `VOXTYPE_OSD_MAX_DURATION_SECS` and the quickshell launcher resolves env > config into `quickshell-style.json`. No more regex, CLI/env overrides included.
- QML reads `style.config.max_duration_secs` (60 fallback). One iteration needed here: the first version synced imperatively on style-object change, but `StyleLoader` swaps only its inner `config` (async FileView load), so the late value never propagated - caught live when a 20s limit still counted from 60 while cutting at 40. It's now a declarative binding, verified counting from 20 and back to 60.

Tested on Omarchy Quattro with a local vulkan build: 60 -> 90 -> 60 round-trips correctly in the runtime JSON, OSD healthy, no QML errors.

OldJobobo (changes requested):
Thanks for the update. I tested the revised implementation, and the effective-limit plumbing works correctly with config files, environment variables, CLI overrides, and an alternate config path. The 71 relevant Rust tests pass, along with Clippy, default binary builds, and GTK4/native frontend compile checks.

I found two remaining countdown correctness issues:

- **Short limits:** QML still applies `Math.max(5, ...)`. With a three-second limit, the daemon stops correctly, but the display still shows **2s remaining**. Could we preserve the effective value without independently clamping it?
- **Restart/reconnect timing:** The countdown measures local timer ticks rather than the recording’s elapsed time. Restarting the OSD halfway through a six-second recording resets the countdown; the daemon stops on schedule while the display still shows **3s remaining**. An audio-bridge disconnect also resets elapsed time, and delayed timer callbacks cause drift. Could we keep the countdown tied to the recording’s timing rather than frontend uptime or meter resets?

These were reproduced with isolated synthetic-audio tests, not just inferred from the code.

Also, `cargo fmt --all -- --check` fails on the new `match` block in `voxtype_osd_quickshell.rs`.

The feature is useful, and the original configuration concern is resolved. I’d like these timing issues addressed before approval so the countdown reliably reflects the actual cutoff.

PR #746, opened 2026-09-10
fix: smooth waveform scrolling in GTK4 and egui OSDs

crmne:
Hey Pete!

I've been working on some new DJ software that I'll announce at some point, and learned a lot getting its waveform display really smooth. I applied those lessons to VoxType's OSD, and both the GTK4 and native egui versions now scroll smoothly with my normal dictation keybinding.

The native frontend requested Wayland frame callbacks but ignored them, redrawing on a separate 16 ms timer. GTK4 also used a timer and only redrew when new audio had arrived. Both moved through the audio history in whole-frame steps, so capture batches showed up as jumps in the waveform.

This change:

- Paces native redraws with Wayland frame callbacks and GTK4 redraws with its frame clock, including between audio arrivals.
- Scrolls a fixed time window using elapsed time and fractional interpolation, with a 50 ms buffer to absorb ordinary capture batches. The window fills with silence at the start instead of stretching the available history across its width.
- Keeps enough history for the configured window, resets it when the audio sequence restarts, and preserves transients when several audio frames fall into one pixel column.
- Draws the egui waveform with triangle strips and antialiased edges, fixing the use of a convex-polygon fill for a concave waveform.

Tested both frontends on my 144 Hz display, including repeated recordings through the usual keybinding. A synthetic test with audio arriving in 40 ms batches measured 7 ms median and 95th-percentile frame-callback intervals in both frontends, and confirmed idle hiding and restart. Unit tests cover timing at 60, 144, and 240 Hz, interpolation, transient preservation, counter wrap, and input stalls.

Please try it. It's smooooth.

Validation: `cargo fmt --check`, default and OSD-feature Clippy with `-D warnings`, builds of both frontends, and `cargo test` (1,120 passed, 2 ignored).

### DIFF
--- a/src/bin/voxtype_osd_gtk4.rs
+++ b/src/bin/voxtype_osd_gtk4.rs
@@ -5,9 +5,8 @@
 //! scrolling waveform plus a segmented peak meter. Audio frames arrive on
 //! the daemon's audio Unix socket via [`voxtype::osd::ipc::run_ipc_loop`],
 //! decoded into [`AudioFrame`]s by a tokio runtime on a worker thread, and
-//! pushed into a shared [`FrameRing`] + [`PeakHold`]. The GTK side polls a
-//! ~60 Hz `glib::timeout_add_local` callback that redraws the
-//! `DrawingArea` whenever new frames have arrived.
+//! pushed into a shared [`FrameRing`] + [`PeakHold`]. The GTK frame clock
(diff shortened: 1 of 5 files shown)


peteonrails:
@crmne This is awesome! Thank you for applying the learning here! Giving it a try

peteonrails (approved):

PR #713, opened 2026-09-03
Add Volcengine Seed-ASR as a native streaming engine

erning:
## Description

Add Volcengine Seed-ASR as a built-in cloud transcription engine.

This change:

- implements the Seed-ASR 2.0 bidirectional binary WebSocket protocol;
- supports new-console API key authentication;
- supports legacy App ID and access token authentication;
- provides native streaming and buffered transcription;
- reconciles cumulative recognition results with the existing streaming event
  contract;
- exposes Seed-ASR through config commands, the configuration schema, and the
  interactive configure form;
- publishes completion outcomes for streamed file output;
- documents configuration, authentication, protocol behavior, and current
  limitations.

## Related Issue

Closes #712

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Testing

- [x] I have tested these changes locally
- [x] I have run `cargo test` and all tests pass
- [x] I have run `cargo clippy -- -D warnings` with no warnings
- [x] I have run `cargo fmt`

Testing included:

- the full unfiltered test suite on Linux;
- the local Seed-ASR WebSocket round-trip test;
- authentication, framing, response parsing, and reconciliation tests;
- configuration schema and interactive configure tests;
- live Seed-ASR recording on macOS.

## Documentation

- [x] I have updated documentation as needed
- [ ] No documentation changes are needed

## Pre-merge checklist

- [x] This PR targets the `dev` branch (not `main`)
- [ ] Wait for `ci-success` to pass before marking Ready for Review (draft mode is encouraged while CI is red)

## Additional Notes

Live Volcengine service tests require account credentials and are not part of
the automated test suite.

PR #758, opened 2026-09-12
Fix stalled Seed-ASR live text after a provisional revision

### LINKED ISSUE #756: Seed-ASR live typing stalls when a provisional hypothesis changes
## Description

I've been testing the Seed-ASR branch from #713 for Chinese dictation on Hyprland. With `type_partials = true`, text starts appearing, but an ordinary correction from the recognizer can make the visible text stop updating until the utterance becomes definite.

This is about the proposed Seed-ASR engine, not the released default engine. The branch I tested is at `303702f`.

## Steps to reproduce

Enable Seed-ASR streaming with `type_partials = true`. The problem can be reproduced without a microphone or a cloud account by feeding these cumulative snapshots to `Reconciler::process`:

1. `text = "这个云书法"`, no stable prefix, not the last packet.
2. `text = "这个语音输入法"`, still no stable prefix, not the last packet.

The second hypothesis corrects the first, but it does not start with the already-typed partial. The current reconciler emits no update for it. Subsequent hypotheses that extend the corrected text remain blocked for the same reason.

There is a related case when a later snapshot changes punctuation in a prefix previously marked definite: the reconciler returns `server revised text that was already finalized`, ending the session.

## Expected behavior

In opt-in live-typing mode, reconcile the latest cumulative hypothesis with the text rendered by this dictation session: keep the shared prefix, remove the divergent suffix, and insert the replacement. A normal provisional correction should not have to wait for definite metadata.

The default, stable-only mode can keep its current commit policy.

## Environment

- Arch Linux / Omarchy, Hyprland, Wayland
- Voxtype built from #713 at `303702f` (reports 1.0.1)
- Native Seed-ASR streaming, `type_partials = true`
- Observed while testing Chinese and mixed Chinese/English input

## Additional context

I'm preparing a focused patch with synthetic regression cases. No account credentials or speech recordings are needed for those tests.

I also encountered stop/drain and keyboard-injection issues during the same testing. Those are separate from this reconciler bug; I'll describe them on #713 rather than fold them into this issue.

sanxi33:
## Description

This is a follow-up to #713. Thanks to @erning for the native Seed-ASR engine; I've been using it for Chinese dictation on Hyprland, and it has made a real difference to my setup.

With `type_partials = true`, a correction such as `这个云书法` → `这个语音输入法` can leave the first version on screen until the utterance becomes definite. The recognizer has already corrected itself, but the new hypothesis no longer extends the old one, so the reconciler suppresses it. A later correction to an already-definite prefix can end the session altogether.

This patch gives live mode a full-transcript `Snapshot` event. The output session compares each snapshot with the text it has rendered, keeps the common prefix, and replaces only the divergent suffix. Empty acknowledgement/final packets retain the last nonempty hypothesis. File output replaces its accumulated text without counting it as keyboard input.

The default stable-only mode is unchanged. Existing output hooks are retained; in live mode the post-processor runs once on the complete final transcript. A failed output transaction stops further rewrites rather than guessing what is left on screen.

**Dependency:** #713 is still open, so this branch contains its five commits unchanged. My patch is just [d30148b](https://github.com/sanxi33/voxtype/commit/d30148ba4d1d020cee8bee153a3a84ac9b886cc9), touching six files. This is a draft against `dev`, not a replacement for #713; the single commit can also be cherry-picked into that branch.

## Related issue

Fixes #756.

## Testing

- `cargo fmt --all --check`
- `cargo clippy --release --offline --all-targets --no-deps -- -D warnings`
- Full release-profile test suite: 1,149 passed, 2 ignored, none failed.
- Added synthetic cases for provisional revisions, changed stability metadata, unchanged/empty finals, Unicode suffix edits, file-output bookkeeping, failed output, and final-only post-processing.

Checks ran on Linux x86_64. The release checks used four build jobs, LTO disabled, and 16 codegen units. No cloud credentials or recordings are required by the tests. I have not tested this patch on macOS.

## Scope

Manual dictation testing was on a broader local build, which also contains focus, audio-drain, and keyboard fixes. For this extracted patch, the reproducible coverage is the automated suite above.

I kept the stop behavior, compositor-specific caret probe, request tuning, and keyboard pacing out of this PR. #494 already addresses a related backspace-pacing problem. I'd like to discuss the focus and drain behavior separately before proposing changes to it here.

## Pre-merge checklist

- [x] Targets `dev`
- [x] Regression tests and Seed-ASR documentation updated
- [ ] Resolve dependency on #713
- [ ] Wait for `ci-success` before marking ready for review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6116,6 +6116,7 @@ dependencies = [
  "egui",
  "egui-wgpu",
  "evdev",
+ "flate2",
  "futures-util",
  "glib 0.22.7",
  "gtk4",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,9 +55,10 @@ hound = "3"  # WAV file reading/writing
 # HTTP client for remote transcription
 ureq = { version = "2", features = ["json"] }
 
-# WebSocket client for Soniox streaming backend
+# WebSocket client for cloud streaming backends
 tokio-tungstenite = { version = "0.24", features = ["rustls-tls-webpki-roots"] }
 futures-util = "0.3"
+flate2 = "1"
 # Async HTTP client for Soniox async transcription API (multipart upload + poll)
 reqwest = { version = "0.12", default-features = false, features = ["json", "multipart", "rustls-tls"] }
 
--- a/config/default.toml
+++ b/config/default.toml
@@ -143,6 +143,22 @@ on_demand_loading = false
 #
 # on_demand_loading = false
 
+# [seedasr]
+# Volcengine Seed-ASR cloud WebSocket engine (engine = "seedasr").
+# Use either api_key or the legacy app_id + access_token pair.
+# api_key = "your-api-key"             # Or SEEDASR_API_KEY
+# app_id = "your-app-id"               # Or SEEDASR_APP_ID
+# access_token = "your-access-token"   # Or SEEDASR_ACCESS_TOKEN
+# resource_id = "volc.seedasr.sauc.duration"
+# url = "wss://openspeech.bytedance.com/api/v3/sauc/bigmodel_async"
+# streaming = true
+# type_partials = false
+# language = "auto"
+# enable_itn = true
+# enable_punc = true
+# enable_ddc = false
+# end_window_ms = 800
+
 [output]
 # Primary output mode: "type", "clipboard", or "paste"
 # - type: Simulates keyboard input at cursor position (requires wtype or ydotool)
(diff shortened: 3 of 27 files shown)


sanxi33:
Closing this for now. Longer real-world dictation is still producing repeated text and incorrect deletions in the broader local setup. The earlier test results were not enough to establish that the approach is reliable, and I do not want this patch to be treated as a ready solution. I will revisit the design and testing before proposing another fix.

PR #770, opened 2026-09-17
feat(output): per-app auto-submit rules

adambuchweitz:
## What

`auto_submit` is currently global. This adds `[output.auto_submit_apps]`, a map of window pattern to auto-submit override, so dictation can press Enter in chat apps and editors without sending stray Enters into markdown files or terminals.

```toml
[output.auto_submit_apps]
Slack = true
discord = true
kitty = false   # terminal running neovim: never submit
```

## How it works

- Patterns are case-insensitive substrings matched against the focused window's class (or title). The longest matching pattern wins, so `neovim` beats `kitty`. No match falls back to the global `auto_submit` setting. An empty pattern `""` matches every window and acts as a catch-all.
- The focused window is queried at output time: `hyprctl -j activewindow` (Hyprland), `swaymsg -t get_tree -r` (Sway), `niri msg --json focused-window` (Niri), or the frontmost app name via osascript on macOS. XWayland windows under Sway fall back to `window_properties.class` when `app_id` is null.
- The compositor query is skipped entirely when no rules are configured, so users without them pay nothing. A 750ms timeout guards against a wedged compositor socket stalling the output path; on timeout or failure the global setting applies.
- Applied in both the normal transcription path and the streaming chain build (streaming drivers bake `auto_submit` in at chain construction, so the rule resolves at session start). Resolution happens after the `--auto-submit/--no-auto-submit` CLI overrides and the smart auto-submit keyword, so those still win.

## Notes

- Not wired into `voxtype config get/set` or the TUI: the schema registry only handles flat scalar keys, and `driver_order`-style non-scalar options are the precedent for TOML-only configuration. Happy to add a scalar representation if maintainers want TUI editing.
- No new dependencies (serde_json, tokio process, and existing infra only).

## Testing

- Unit tests for pattern matching, resolution precedence, TOML parsing, and the compositor JSON parsers (Hyprland/Sway tree walk/Niri/XWayland fallback). `cargo test` passes (1149 tests).
- Verified live on Hyprland: focused-window query, rule resolution, and Enter-after-dictation in matched apps.

Note for reviewers: the Sway/Niri/macOS paths are covered by unit tests of the JSON parsing layer but have not been exercised on those compositors.


adambuchweitz:
Sorry about the churn on this. Accidentally targeted the RC branch instead of dev

